PR #78, opened 2026-07-13
fix: every relay converges to the latest match state

grunch:
## Problem

While refereeing, updates would sometimes never reach the relay being monitored. It happened rarely when adding points, but often when adding advantages or penalties in quick succession — three advantages could go by with nothing published, and only some later action would eventually carry the accumulated state across.

## Root cause

`publishEvent` fans out to all connected relays but treats **"at least one relay accepted"** as success. Once any relay sends OK, the outbox is cleared and that state is **never resent to the other relays**.

So if the relay you watch rejects the event (rate-limiting a burst of rapid taps — advantages and penalties get hit in rapid succession far more often than points) or is disconnected at that moment, while the other relay accepts, the app considers the publish done and that relay keeps serving a stale scoreboard. A later action, landing when the connection is calm, reaches both relays and drags the accumulated state along — which is exactly the "occasionally a future event catches up" behavior observed.

## Fix

`NostrService` now guarantees **every configured relay** converges to the latest state of each match:

- **Per-relay tracking** (`_pendingLatest` / `_pendingAcks`): on publish, the latest signed event per match (d-tag) is registered along with which relays OK'd it. Publish still resolves on the first OK so the UI never stalls — but the work doesn't stop there.
- **Resend on reconnect**: when a relay (re)connects, it immediately receives the latest event it's missing.
- **Resend sweep** (every 5s while any relay is behind): connected relays that rejected the event get it again until they accept.
- **Only the newest state travels**: if three updates happened while a relay was behind, only the last one is resent — addressable events supersede their predecessors, so intermediate states are noise. An in-flight guard prevents duplicating a frame already awaiting OK on the same relay.

Transport hardening underneath:

- A relay that never answers OK is treated as a **dead socket** and recycled, instead of being trusted for every subsequent publish. (Backgrounding/network switches kill sockets without a close frame, so `isConnected` can lie for minutes.)
- All relay connections are recycled **on app resume**.
- A pending match state is published **the moment a relay reconnects**, instead of waiting out a backoff scheduled against a socket that's already dead.

## Test plan

- [x] `a relay that missed an event receives it after reconnecting`
- [x] `a relay that rejects an event gets it resent until accepted` (fails on `main` — this is the reported bug)
- [x] `only the newest state is resent when updates supersede each other`
- [x] `publish drops the connection when the relay never answers`
- [x] `reconnectNow tears down the socket and connects a fresh one`
- [x] `a pending state is published the moment a relay reconnects`
- [x] `a reconnect with nothing pending publishes nothing`
- [x] Full suite green: 105 tests
- [x] `flutter analyze`: no new issues
- [ ] Manual: create a match, hammer the advantage button, confirm every press lands on the monitored relay in real time

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relay reliability by detecting stalled connections and automatically reconnecting.
  * Ensured the latest match state is retried and synchronized across relays after reconnections or temporary failures.
  * Pending updates now publish promptly when connectivity is restored.

* **Improvements**
  * The app automatically reconnects to relays when returning to the foreground.
  * Prevented updates from continuing after related components are closed.

### DIFF
--- a/lib/features/match/providers/match_control_provider.dart
+++ b/lib/features/match/providers/match_control_provider.dart
@@ -74,12 +74,24 @@ class MatchControlNotifier extends StateNotifier<MatchControlState> {
   bool _sending = false;
   Timer? _retryTimer;
   int _retryAttempt = 0;
+  StreamSubscription<String>? _relayConnectedSub;
 
   MatchControlNotifier(Match match, this._nostrService, [this._feedNotifier])
       : super(MatchControlState(
           match: match,
           remainingSeconds: _calculateRemaining(match),
         )) {
+    // A relay that just (re)connected can take the pending state right now —
+    // waiting out a backoff scheduled against its dead predecessor would
+    // leave the remote board stale for up to 30 extra seconds.
+    _relayConnectedSub = _nostrService.onRelayConnected.listen((_) {
+      if (_outbox == null) return;
+      _retryTimer?.cancel();
+      _retryTimer = null;
+      _retryAttempt = 0;
+      _drainOutbox();
+    });
+
     if (match.status == MatchStatus.inProgress) {
       if (match.pausedAt != null) {
         // A paused clock stays paused, however long the app was away.
@@ -393,6 +405,7 @@ class MatchControlNotifier extends StateNotifier<MatchControlState> {
   void dispose() {
     _timer?.cancel();
     _retryTimer?.cancel();
+    _relayConnectedSub?.cancel();
     super.dispose();
   }
 }
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -76,11 +76,41 @@ void main() async {
 /// Watches [localeProvider] and [themeModeProvider] to configure the app's
 /// locale and theme mode. Provides both light and dark themes, with the
 /// active mode determined by user preference or system setting.
-class ChokeApp extends ConsumerWidget {
+///
+/// Also watches the app lifecycle: relay sockets rarely survive
+/// backgrounding — the OS kills them without a close frame, leaving
+/// connections that look open but drop every event — so on resume all
+/// relay connections are recycled before the operator's next action.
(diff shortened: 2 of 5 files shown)


grunch:
Thanks — all three findings were real and are fixed in c5f481f.

**1. Stale socket callbacks after `reconnectNow()`** (codex P2 / coderabbit major) — confirmed. `connect()` never held the `stream.listen` subscription, so a late `onDone` from the socket that `reconnectNow()` had already replaced would run `_handleDisconnect()` against the **fresh** `_channel`, close it, and schedule another reconnect.

**2. `_awaitingOk` not scoped to the live socket** (codex P2) — confirmed, and it compounded #1: a publish still awaiting an OK kept its event id marked in-flight after the swap, so the reconnect-triggered resend skipped it as a duplicate. The latest match state stayed stale until `okTimeout` fired, which then disconnected the healthy new socket.

Both come from the same gap — nothing ever *detached* the old socket. Teardown is now explicit (`_teardownChannel()`): cancel the listener, close the sink, and fail every publish waiting on an OK that can no longer arrive, which frees those events for resend on the new socket. `_awaitingOk` (a bare `Set<String>`) became `_inFlight` (`Map<String, Completer>`) so the pending publishes are actually reachable at teardown; `disconnect()` and `_handleDisconnect()` route through the same path.

**3. Test fake reused across reconnects** (coderabbit minor) — confirmed. The `channelFactory` now mints a fresh `_FakeWebSocketChannel` per `connect()`, as production does, so the reconnect path is genuinely exercised instead of re-listening to a dead single-subscription stream.

Two regression tests pin the fixes (both fail on the previous commit):
- `a late close from the old socket does not kill the new connection`
- `reconnecting frees a publish that was in flight on the old socket`

Suite: 107 tests green; `flutter analyze` clean on the touched file.

PR #82, opened 2026-07-13
refactor(nostr): Phase 2 — a NostrCrypto seam between the app and its crypto library

grunch:
Phase 2 of [the migration spec](docs/specs/nostr-sdk-migration.md). The app no longer talks to a crypto library — it talks to a `NostrCrypto` interface. Swapping the implementation in Phase 4 becomes a change to **one line of `main.dart`**.

## What's here

- **`NostrCrypto`** (+ `UnsignedNostrEvent`), and **`NostrToolsCrypto`** implementing it over `nostr_tools`. `KeyManager` and `NostrService` take it by constructor; `main.dart` is the single place that picks the implementation.
- **`package:nostr_tools` is now imported in exactly one file.** `NostrEvent`'s `toNostrToolsEvent`/`fromNostrToolsEvent` converters moved *inside* `NostrToolsCrypto` — the event model is now free of any crypto library's types, which is what the acceptance criterion actually requires.
- The dead `nostr_tools` import in `match.dart` is gone.
- **A contract test suite written against the interface, not the implementation.** Phase 3 will hold the Rust backend to these same 19 tests without editing a line of them, so "the new backend behaves like the old one" stops being a claim and becomes a test result.

### 1. `nostr_tools` under-verifies events — a real flaw (§2.1)

Writing the contract first paid for itself immediately. `EventApi.verifySignature` checks the signature **against the event id, but never checks that the id describes the event**. Per NIP-01 the id *is* the hash of the event's fields — so an event whose content is rewritten after signing, keeping the original id and signature, is forged. **`nostr_tools` calls it valid.** The Rust `nostr` crate rejects it (Phase 1's tampering test already proved that), so the two implementations were not actually interchangeable.

`NostrToolsCrypto.verifyEvent` now recomputes the id (`getEventHash`) and rejects a mismatch. This is a deliberate, small hardening rather than a pure move — it's what makes the contract satisfiable by both backends. **It cannot regress the app**, whose only use of verification is a self-check on events it has just signed itself (where the id always matches by construction). Exposure was nil, but it is precisely the class of bug that motivated this migration in the first place.

### 2. Invariant I7 was never actually pinned (§2.2)

The spec claimed I7 ("same nsec always derives the same npub; import round-trips") was *"pinned by key manager tests"*. **There were no key manager tests.** Since Phase 4 swaps the crypto backend underneath exactly that code, the invariant needed teeth *before* then, not after. Added: first launch, reopening the app, npub/nsec round-trips, nsec import (valid and malformed), and the repair path for a stored pubkey that disagrees with its private key — all against an in-memory secure storage.

## Test plan

- [x] `flutter test` — **136 passing** (110 existing + 19 contract + 7 key manager)
- [x] `flutter analyze` — 51 issues, one *fewer* than `main` (the dead import); zero introduced
- [x] Android APK builds with the seam in place
- [x] `grep -rn "package:nostr_tools" lib/` returns exactly one file
- [x] Behavior-preserving apart from the verification hardening in §2.1

## Rollback

Revert the PR. Nothing changes shape for the caller: `KeyManager()` and `NostrService(keyManager)` still work with no arguments, defaulting to `NostrToolsCrypto`.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated cryptography layer for key generation, NIP-19 encoding/decoding, event signing, and verification.
  - Added stronger protection against tampered or forged events.
  - Key management now supports reliable identity recovery, import, validation, and repair.

- **Refactor**
  - Decoupled application Nostr functionality from its underlying cryptography implementation.

- **Tests**
  - Added comprehensive coverage for key persistence, imports, round-trips, event signing, and signature validation.

- **Documentation**
  - Updated migration progress, contract findings, invariants, and acceptance criteria.

### DIFF
--- a/docs/specs/nostr-sdk-migration.md
+++ b/docs/specs/nostr-sdk-migration.md
@@ -1,12 +1,12 @@
 # Migration Spec: `nostr_tools` (Dart) → `nostr-sdk` (Rust) via `flutter_rust_bridge`
 
-**Status:** APPROVED — Phases 0-1 complete, Phase 2 next
+**Status:** APPROVED — Phases 0-2 complete, Phase 3 next
 **Author:** prepared with Claude Code
 **Date:** 2026-07-13
 **Decisions locked (2026-07-13):** web target frozen (§5, W1) · own thin Rust
 crate over official Flutter bindings (§9.2) · manual QA on Android (§9.3)
 
-**Progress:** Phase 0 ✅ · Phase 1 ✅ · Phases 2–8 pending
+**Progress:** Phases 0 ✅ 1 ✅ 2 ✅ · Phases 3–8 pending
 
 ---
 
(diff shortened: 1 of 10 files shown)


grunch:
@coderabbitai review

grunch:
@codex review

PR #83, opened 2026-07-13
feat(rust): Phase 3 — Rust crypto implementation, proven equivalent by differential tests

grunch:
Phase 3 of [the migration spec](docs/specs/nostr-sdk-migration.md). Both crypto backends now exist side by side. **The default is still `nostr_tools`** — Phase 4 flips it. This PR's job is to earn that flip.

## What's here

- **The crate gains keys, NIP-19 and signing**: `generate_secret_key`, `public_key_from_secret`, `npub_encode`, `nsec_encode`, `nsec_decode`, `finish_event`, `verify_event_data`. Ten Rust unit tests, including the NIP-19 spec vector.
- **`RustNostrCrypto`** implements the Phase 2 `NostrCrypto` interface over the generated bindings.
- **Backend selector**: `--dart-define=NOSTR_BACKEND=rust|legacy`, default `legacy`. `main.dart` initializes `RustLib` only when Rust is selected — so rolling back Phase 4 will need no code change at all.

## A design decision worth flagging: structs, not JSON, across the bridge (§3.1)

The event id is a **hash of the event's canonical serialization**. Shipping events over FFI as JSON would have put a *second* serializer in the path — Dart's `jsonEncode` on one side, `serde` on the other. That's a silent canonicalization the differential tests could only have caught by luck. So the bridge carries the **fields** (`UnsignedEventData` / `SignedEventData`), and the `nostr` crate performs the only serialization that counts.

## The proof (§3.2)

The Rust implementation passes the **identical Phase 2 contract — 19 tests, not a line changed** — plus 10 differential tests. The two libraries agree on:

- **public key, npub and nsec** from the same private key (20 random keys per run), and each accepts the other's generated keys;
- **the event id**, for every hostile content vector tried — empty, accented (`Gonçalves`), CJK (`田中太郎`), emoji (`🥋⚡`), embedded quotes, newlines, tabs, back- and forward-slashes, raw match JSON — and for tag sets from empty to unicode d-tags. **Identical ids mean both libraries serialize to identical canonical bytes**, which is the whole ballgame;
- **`created_at`** preserved exactly across the FFI boundary (0 → 4102444800) — this matters because NIP-01 replacement ordering turns on it (I1), and a timestamp mangled in transit would silently break the convergence work from #78;
- **rejecting a post-signing tamper**, keeping the divergence found in §2.1 closed.

Signatures are deliberately **not** compared byte for byte — Schnorr signing uses a random nonce, so they legitimately differ every time. What's asserted instead is the property that actually matters: **each library verifies the other's signatures**, which is exactly what a relay and every other Nostr client will do.

## Size, re-measured

§0.3 asked Phase 3 not to assume headroom, so:

| Release APK (arm64) | Size |
|---|---|
| Baseline (`main` before Phase 1) | 41.7 MB |
| This PR | 43.9 MB |
| **Delta** | **+2.2 MB** |

The seven new crypto functions cost only ~0.08 MB on top of Phase 1 — secp256k1 was already paid for. Budget is ≤6 MB.

## Test plan

- [x] `cargo test` — 10 passing; `cargo clippy -D warnings` and `cargo fmt --check` clean
- [x] `flutter test` — **165 passing**
- [x] `flutter test --tags rust` — 29 passing (19 contract + 10 differential)
- [x] `flutter analyze` — 51 issues, unchanged from `main`
- [x] APK builds on **both** backends (default, and `--dart-define=NOSTR_BACKEND=rust`)
- [ ] Manual QA on Android with the Rust backend — deferred to Phase 4, where it becomes the default and the checklist (same npub after upgrade, nsec import, full match lifecycle against a real relay) is the acceptance gate

## Rollback

Revert the PR, or simply don't pass the flag: the default path is untouched `nostr_tools`.

## Summary by CodeRabbit

* **New Features**
  * Added an optional Rust-powered Nostr crypto backend, selectable at app startup.
  * Added key generation, public/private key conversion, NIP-19 encoding and decoding, event signing, and signature verification.
  * Enabled typed event data exchange between the app and native crypto layer.

* **Tests**
  * Added cross-backend validation for keys, encodings, event IDs, signatures, tampering detection, and timestamp preservation.

* **Documentation**
  * Updated migration progress through Phase 3 and documented interoperability results and acceptance criteria.

### DIFF
--- a/docs/specs/nostr-sdk-migration.md
+++ b/docs/specs/nostr-sdk-migration.md
@@ -1,12 +1,12 @@
 # Migration Spec: `nostr_tools` (Dart) → `nostr-sdk` (Rust) via `flutter_rust_bridge`
 
-**Status:** APPROVED — Phases 0-2 complete, Phase 3 next
+**Status:** APPROVED — Phases 0-3 complete, Phase 4 next
 **Author:** prepared with Claude Code
 **Date:** 2026-07-13
 **Decisions locked (2026-07-13):** web target frozen (§5, W1) · own thin Rust
 crate over official Flutter bindings (§9.2) · manual QA on Android (§9.3)
 
-**Progress:** Phases 0 ✅ 1 ✅ 2 ✅ · Phases 3–8 pending
+**Progress:** Phases 0 ✅ 1 ✅ 2 ✅ 3 ✅ · Phases 4–8 pending
 
 ---
 
@@ -320,7 +320,7 @@ the invariant needed teeth before then, not after.
 
 ---
 
-### Phase 3 — Rust crypto implementation + differential tests *(PR: additive)*
+### Phase 3 — Rust crypto implementation + differential tests — ✅ **DONE (2026-07-13)** *(PR: additive)*
 
 **Goal:** a fully verified Rust implementation, not yet default.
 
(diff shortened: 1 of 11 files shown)


grunch:
@coderabbitai review

PR #85, opened 2026-07-13
refactor(nostr): Phase 5 — a NostrRelayBackend seam under the relay layer

grunch:
Phase 5 of [the migration spec](docs/specs/nostr-sdk-migration.md). **Pure refactor** — the transport becomes swappable and nothing about the app's behavior changes. Independent of Phase 4 (#84), so it branches from `main` and the two can land in either order.

## The split, and why it's the whole point (§5.1)

The interface is deliberately narrow: connect, subscribe, publish to **one** relay, report what **that relay** said. Everything that makes publishing *trustworthy* stayed above it in `NostrService`:

- the outbox and per-relay convergence, with its ack bookkeeping
- the resend sweep, and supersession (only the newest state per match is resent)
- monotonic `created_at` (NIP-01 replacement ordering)
- the addressable cache and NIP-40 expiry filtering

Each of those was **bought with a production bug** (#77, #78), and none is a property of a *transport*. Had they been pushed down into the backend, Phase 6 would have to re-implement and re-prove every one of them against `nostr-sdk`'s relay pool — and the bug that started this whole thread, *a relay quietly serving a stale scoreboard*, would have had a fresh place to hide.

What the transport genuinely does own, because it really is about sockets: the **OK-confirmed publish** (a relay's "sent" is not its "accepted"), and treating a **silent relay as dead, not slow**.

## What moved

`RelayConnection` moved **verbatim** into `DartRelayBackend` — its subtleties intact — and its tests moved with it into `relay/dart_relay_backend_test.dart`, where they now read as what they actually are: transport regressions, not service tests.

## Test plan

- [x] `flutter test` — **165 passing**, zero behavior change
- [x] `flutter analyze` — 51 issues, unchanged
- [x] **Acceptance:** `NostrService` no longer imports `web_socket_channel` — it cannot form an opinion about sockets any more

## Rollback

Revert the PR. Pure refactor; the public surface of `NostrService` is unchanged, and `NostrService(keyManager)` still works with no backend argument (defaulting to `DartRelayBackend`).

## Summary by CodeRabbit

- **New Features**
  - Improved relay connectivity and event publishing across multiple Nostr relays.
  - Added more reliable reconnection, subscription management, and publishing status handling.

- **Bug Fixes**
  - Prevented stale connections from interfering after reconnecting.
  - Improved timeout handling and retry behavior for unacknowledged events.

- **Documentation**
  - Updated migration documentation with completed phases, architecture progress, and validation results.


grunch:
@codex review

grunch:
@coderabbitai review

grunch:
Both findings were real and are fixed. They're in code this PR *moved* rather than wrote — but it ships for a release cycle as the rollback path, so they'd have shipped with it.

## 1. A connect still shaking hands could hijack its replacement (Major)

Confirmed. `connect()` dialed a socket, awaited the handshake, then **read the channel field back** — by which point it might be somebody else's. The app calls `reconnectNow()` on resume, and a slow relay is easily still shaking hands when that lands.

The part that actually bites isn't the double listener — it's the **spurious "connected" announcement**. `NostrService` resends a match's pending state on every connection, so it would push the score to a relay on the strength of a handshake belonging to a socket that no longer exists.

An attempt now carries the generation it was born in and does nothing if it's no longer current; teardown bumps the generation, so anything in flight is orphaned by construction. The regression test asserts on exactly that: **with the bug the relay announces two connections where there was one** (`Expected: <1> Actual: <2>`).

Also fixed alongside it: the generic error branch left its failed socket **open** — only the timeout branch closed it.

## 2. Raw relay frames in the logs (Major)

Confirmed. Frames carry match content and pubkeys, and `debugPrint` is **not** stripped from release builds. Only the frame's size is logged now.

## A bug I introduced fixing #1, and how it got caught

My first version `await`ed `sink.close()` inside `connect()`'s catch block. Closing a socket that never opened can throw — and that exception **escaped the catch, so `_scheduleReconnect()` never ran**. A relay that went down would never come back: precisely the failure this class exists to prevent.

It was caught by the convergence drills on the Phase 7 branch (the three that depend on a relay returning went red). Closing is best-effort now — never awaited, never allowed to throw. Worth noting that the first version of the regression test *passed without the fix*, which is why I checked it actually goes red first; the assertion was rewritten to target the real harm.

Green across the stack: **166** (this PR) / 186 (#86) / 194 (#87) / 142 (#88), analyzer unchanged. The fix is propagated through all four branches.

grunch:
Went back over the review to confirm nothing was left. There is **one review, with 2 actionable comments and 3 nitpicks** — CodeRabbit has not re-reviewed since the push. All five are now addressed.

## The 2 actionable ones — fixed earlier (`fdcf2b1`, `49f34f5`)

1. **A connect still shaking hands could hijack its replacement.** Attempts now carry the generation they were born in; teardown bumps it, so anything in flight is orphaned by construction. The regression test asserts on the harm that actually matters — the spurious "connected" announcement, which makes `NostrService` push a score to a relay on the strength of a dead socket's handshake. Also: the generic error branch was leaving its failed socket open.
2. **Raw relay frames in the logs.** Frames carry match content and pubkeys, and `debugPrint` is not stripped from release builds. Only the frame's size is logged now.

## The 3 nitpicks — I had not read them. Now handled

3. **`const Filter`** — done. Its fields are final and it has no body; there was no reason for it not to be.
4. **`eventStream` / `dispose()` undocumented** — done.
5. **"Document the complete public relay API surface"** — done for the parts that live: `Filter`'s fields, constructor and `toJson()` (including *why* unset fields are omitted rather than sent as null — a relay reads an explicit null as a constraint nothing can satisfy), and the interface's lifecycle methods.

   **Skipped** for `RelayConnection` and `DartRelayBackend`'s members, which the same nitpick also asked for: **Phase 8 (#88) deletes both files.** Doc comments on code already scheduled to go are churn a reviewer has to read twice.

## While propagating, `main` had moved

Phase 4 (#84) merged in the meantime, so this branch picked it up. That produced a real conflict in `main.dart` — Phase 4 flips the crypto default to `rust`, Phase 6 adds the relay selector — resolved to keep both. The whole stack was rebuilt and re-run:

| Branch | Tests |
|---|---|
| #85 (this) | **170** |
| #86 | 190 |
| #87 | 198 |
| #88 | 142 |

All green, `--tags rust` included; analyzer unchanged.

PR #86, opened 2026-07-13
feat(relay): Phase 6 — the nostr-sdk relay backend, proven by a transport contract

grunch:
Phase 6 of [the migration spec](docs/specs/nostr-sdk-migration.md). **Depends on #85** (Phase 5) — it is based on that branch, so review/merge #85 first; this PR targets it rather than `main`.

Both transports now exist side by side. **The default is still the hand-written one** — Phase 7 flips it.

## I dropped shadow mode for something stronger (§6.1)

The spec called for *shadow mode*: run the Rust pool read-only alongside the Dart one and watch for divergence. I replaced it with **a transport contract run against a real relay**, which is both cheaper and far more convincing.

`fake_relay.dart` is an in-process Nostr relay — a real WebSocket server speaking NIP-01 (`EVENT` → `OK`, `REQ` → events → `EOSE`), able to accept, to **reject with a reason**, and to push events to subscribers. Both backends are then subjected to the *identical* contract.

The wire is the only place they **can** be compared: they share no code at all (Dart over `web_socket_channel`, Rust over `nostr-sdk`). Shadow mode would only have shown they *tend* to agree in whatever conditions happened to arise. The contract asserts they agree on what the app's reliability actually rests on:

- **`publish` reports what *that relay* said** — accepted, or rejected. Convergence (I3) is built entirely on that distinction; a backend reading "sent" as "accepted" would resurrect #78 on the spot.
- a relay that is up is counted; one that leaves stops being counted
- a connecting relay is announced (the cue to push what it missed)
- events publish and arrive **with every field intact** — content, tags, `created_at` — through UTF-8, emoji, embedded JSON
- publishing to a relay that was never dialed **fails**, instead of quietly pretending

**Both backends pass all ten, unchanged.**

## The contract earned its keep immediately (§6.2)

The subscription tests failed against Rust at first. The reason is worth keeping: **`nostr-sdk` verifies the signature of every event it receives and drops the ones that don't check out.** My fixtures had stub signatures, so Rust — correctly — refused to deliver them.

**The Dart transport had never checked.** It would hand the app any event a relay cared to send, forged or not. That's the same class of gap as §2.1 (where `nostr_tools` accepted events whose content had been rewritten after signing), and it's a **second, unadvertised win** for this migration: after Phase 7, a broken or malicious relay can no longer inject a match that the referee never signed.

The contract now signs its events for real.

## Design notes (§6.3)

- **Publish one relay at a time.** `nostr-sdk` will fan out inside its pool, but its `Output.success`/`.failed` maps are *exactly* the per-relay verdict our convergence bookkeeping needs. Flattening that into a pool-wide "it worked" would throw away the very thing #78 was fixed to preserve.
- **The status stream reports disconnections too**, not just connections. `connectedRelays` is answered synchronously on the hot path (deciding where to publish a score), so Dart keeps its own view — and a stream that only ever said "connected" would leave that view permanently optimistic, publishing into relays that had left.
- **Raw `Message` notifications, not the de-duplicated `Event` ones** — the latter drop events the client sent itself, and the app needs the relay's echo of its own matches: that echo is how a second device learns the score.

## A bug I introduced and caught while wiring

`RustLib.init()` lived inside the crypto builder. Since the two backends are selected **independently**, a build taking the Rust *transport* with the legacy *crypto* would never load the library the transport needs — fine on a laptop, broken on a phone. It's hoisted out now.

## Test plan

- [x] `flutter test` — **185 passing**
- [x] `flutter test --tags rust` — 42 passing (crypto contract ×2, differential, backend swap, **transport contract ×2**)
- [x] `cargo test` — 10; `clippy -D warnings` and `fmt --check` clean
- [x] APK builds with `--dart-define=NOSTR_RELAY_BACKEND=rust`
- [ ] Failure drills against live relays — deferred to Phase 7, where the transport becomes the default and those drills are the acceptance gate

## Rollback

Revert, or simply don't pass the flag: the default path is the unchanged Dart transport.


grunch:
@coderabbitai review

PR #87, opened 2026-07-13
feat(relay): Phase 7 — make nostr-sdk the default transport

grunch:
Phase 7 of [the migration spec](docs/specs/nostr-sdk-migration.md). **Depends on #86** (which depends on #85) — it targets that branch, so the chain merges 85 → 86 → 87.

`nostr-sdk`'s relay pool now carries the app's traffic. Rollback is a flag: `--dart-define=NOSTR_RELAY_BACKEND=legacy`.

## The failure drills are automated, not remembered (§7.1)

The spec listed this phase's drills as **manual**: airplane mode mid-match, kill one relay of two, hammer a rate-limiting relay. Those are exactly the scenarios that produced the original bug — and exactly the ones a human is worst at re-enacting faithfully: on a phone, under time pressure, once, from memory.

They're now `convergence_drills.dart`, run against a **real relay**, on **both** transports:

- **a relay that rejects the score is retried until it accepts** — rate limiting, which is precisely what a burst of advantages triggers (your original report)
- **a relay that was down gets the score when it comes back** — both relays configured, one dead when the referee scores, then returning *on the same address*. This is the whole of #78: "one relay accepted" is not "done", and the relay that was away is exactly the one whose scoreboard goes stale
- **a relay that was down receives only the latest score, not the history** — the events are addressable; replaying three intermediate scores would make a relay briefly display a score that is no longer true
- **with nothing reachable, the score survives until the relay returns** — airplane mode: the publish fails *loudly* (pretending is what got us into this), and the score lands by itself once the network is back, with the referee tapping nothing

**All four pass on both transports.**

Writing them caught **my own first attempt measuring the wrong thing**: it added the second relay *after* the publish, and convergence is measured against the relays configured *at publish time*. The drills now configure both up front and take one down — which is the situation the app is actually in.

## Size went over budget, and was fixed rather than waved through (§7.2)

The first release build came in at **48.2 MB** — a **6.5 MB** delta over the 41.7 MB baseline, **breaking the ≤6 MB ceiling**. `nostr-sdk` brings the relay pool, tokio and TLS (its optional NIPs were already off).

The crate was building with Cargo's **stock release profile**. Optimizing it for size — `opt-level = "z"`, LTO, one codegen unit, stripped — takes the native library from **6.12 MB → 3.57 MB**:

| Release APK (arm64) | Size |
|---|---|
| Baseline | 41.7 MB |
| This PR | **45.5 MB** |
| **Delta** | **+3.8 MB** ✅ |

`panic = "abort"` is deliberately **not** set: flutter_rust_bridge unwinds across the FFI boundary to turn a Rust panic into a Dart exception.

## Test plan

- [x] `flutter test` — **193 passing**
- [x] `flutter test --tags rust` — 46 passing (crypto contract ×2, differential, backend swap, transport contract ×2, **drills ×2**)
- [x] `cargo test` 10; `clippy -D warnings`, `fmt --check` clean
- [x] Release APK builds on the new default; `legacy` flag still builds

## Manual QA still outstanding (needs a device)

- [ ] a full match against live relays, watched on a second Nostr client
- [ ] airplane mode mid-match, on the actual phone
- [ ] app backgrounded ten minutes mid-match, then resumed

The drills cover these behaviors against a real relay, but a real device adds real radios and a real OS killing real sockets — worth one pass before release.

## Rollback

`--dart-define=NOSTR_RELAY_BACKEND=legacy`. `DartRelayBackend` stays for one release cycle, removed in Phase 8.

## Summary by CodeRabbit

- **New Features**
  - Rust is now the default relay backend for improved consistency across supported platforms.
  - Native library initialization now fails early with a clear error when required components are unavailable.
  - Release builds include size optimizations for smaller application packages.

- **Bug Fixes**
  - Improved relay recovery and event synchronization when connections fail or return.
  - Pending updates are now delivered automatically when connectivity is restored.

- **Documentation**
  - Updated migration progress, rollout guidance, and quality checks.

- **Tests**
  - Added automated convergence checks covering both legacy and Rust relay backends.


grunch:
@coderabbitai review

PR #89, opened 2026-07-13
ci: ship an arm64-only release APK (86 MB → 36 MB)

grunch:
The release APK was **86.1 MB**, and more than half of that was native code nobody runs: three copies of everything, one per ABI. Since the Nostr stack became a Rust library, that included three copies of *it* too.

| | Release APK |
|---|---|
| Before | 86.1 MB |
| After | **35.7 MB** |
| | **−59%** |

**The filename is unchanged.** The build still emits `app-release.apk`, so the release artifact is still `choke-vX.Y.Z.apk`. I deliberately did **not** use `--split-per-abi`, which renames the output to `app-arm64-v8a-release.apk` and would have broken the rename step.

## The flag alone wasn't enough

`flutter build apk --release --target-platform android-arm64` is the obvious change, and it's *not sufficient* — I checked what actually ended up in the archive:

- it narrows **Flutter's own** libraries, but leaves **Cargokit** building the Rust crate for every Android ABI;
- it doesn't touch native code that arrives **prebuilt inside a plugin's AAR**. ML Kit's barcode scanner — the QR reader on the Account screen — was shipping **8.8 MB** of `x86_64` and `armeabi-v7a` libraries all by itself.

So the workflow flag is paired with two things in Gradle: `ndk.abiFilters` (reaches Cargokit) and a `jniLibs` exclude (reaches the AARs). Verified by listing the archive, not by assuming:

```
before:  arm64-v8a 24.9 MB · x86_64 27.6 MB · armeabi-v7a 20.2 MB
after:   arm64-v8a 24.9 MB
```

## Debug builds keep every ABI, on purpose

Both filters are gated on the build actually being a release. Applying them to debug would **break an x86_64 emulator at runtime** — an `UnsatisfiedLinkError`, with no hint as to why. Verified: the debug APK still carries all three.

Every 64-bit Android phone of the last decade is `arm64-v8a`, so nothing a referee holds is affected.

## Test plan

- [x] Release APK: **35.7 MB**, contains `arm64-v8a` and nothing else
- [x] Output filename is still `app-release.apk` → artifact still `choke-vX.Y.Z.apk`
- [x] Debug APK still contains all three ABIs (emulators keep working)
- [x] `flutter test` — 142 passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android release builds by targeting arm64 devices.
  * Reduced release package compatibility issues involving Rust-based native libraries.
  * Debug builds continue to support multiple device architectures.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -220,8 +220,16 @@ jobs:
           echo "keyAlias=${{ secrets.KEY_ALIAS }}" >> android/key.properties
           echo "storeFile=choke-upload-key.jks" >> android/key.properties
 
+      # arm64 only. A plain `flutter build apk` packs native code for three ABIs
+      # into one file — and since the Nostr stack became a Rust library, that is
+      # three copies of it. Every 64-bit Android phone made in the last decade
+      # runs arm64-v8a; armeabi-v7a and x86_64 were costing ~40 MB to serve
+      # 32-bit hardware nobody referees on, and emulators.
+      #
+      # Deliberately NOT --split-per-abi: that renames the output to
+      # app-arm64-v8a-release.apk and would break the step below.
       - name: Build APK
-        run: flutter build apk --release
+        run: flutter build apk --release --target-platform android-arm64
 
       - name: Rename APK
         run: |
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -16,6 +16,16 @@ if (keystorePropertiesFile.exists()) {
 val releaseKeystoreFile =
     keystoreProperties["storeFile"]?.toString()?.let { file(it) }
 
+// Strip every ABI but arm64 — but only when building a release artifact.
+//
+// Every 64-bit Android phone of the last decade is arm64-v8a, so nothing a
+// referee holds is affected, and it halves the download. Debug builds keep all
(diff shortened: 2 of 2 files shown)


grunch:
@coderabbitai review

grunch:
Valid, and fixed.

`gradle.startParameter.taskNames` is **invocation-wide**, not per-variant. Any run that built both at once — and `gradle build` does exactly that — would have flipped the filter on for **debug** as well. An x86_64 emulator would then fail at runtime with an `UnsatisfiedLinkError` and **nothing in the build output to explain why**, which is the worst kind of breakage: the build is green and the app just dies.

## The fix

- `abiFilters` moves into `buildTypes.release`, where Gradle scopes it by variant.
- The `jniLibs` excludes move to `androidComponents.onVariants(selector().withBuildType("release"))`.

## Verified on the failing case, not just the happy one

A single Gradle invocation building **both** variants — the exact scenario reported:

```
$ ./gradlew assembleDebug assembleRelease

release → arm64-v8a
debug   → arm64-v8a  armeabi-v7a  x86_64
```

Release is still 35.7 MB with arm64 alone; debug keeps all three, so emulators keep working. (Worth noting my earlier "buildTypes.release doesn't work" conclusion was wrong — it came from a stale APK, because a `flutter clean -q` had silently failed on an invalid flag. It works.)

PR #90, opened 2026-07-13
docs: spec — recording *how* a match was won (submission, DQ, forfeit)

grunch:
Spec only, no code. Ready for review.

## The problem, concretely

Bob is beating Carlos **4–0**. The clock is still running. Carlos catches an armbar and Bob taps.

**Carlos won. The app cannot say so.**

Holding *Finish* sets `status: "finished"` and nothing else. The event published to Nostr says Bob leads 4–0 and the match is over — so every dashboard, every other client, and our own match list will show **Bob as the winner of a match he lost**.

The score isn't merely incomplete. It's **actively wrong**, and nothing in the event contradicts it. And this isn't an edge case: submission is the *point* of the sport, and it's the one victory condition that **ignores the scoreboard entirely**. Same for disqualification — the fighter who was ahead can still lose.

## The proposal

Four additive fields on the kind 31415 content:

```json
{
  "status": "finished",
  "winner": "f2",
  "method": "submission",
  "submission": "armbar",
  "ended_at": 1700000180
}
```

- **`winner`** — `f1` / `f2` / `draw`
- **`method`** — `submission`, `points`, `advantages`, `penalties`, `decision`, `dq`, `forfeit`, `draw`. Three of those (`submission`, `dq`, `forfeit`) **beat the scoreboard**; the rest *are* the scoreboard.
- **`submission`** — the technique, optional **free text**, deliberately not an enum. BJJ invents submissions faster than a spec can enumerate them, and a referee must never be blocked from finishing a match because the app has never heard of a *baratoplata*.
- **`ended_at`** — not derivable: `start_at + duration` is when the clock *would* have run out, which is precisely what a submission prevents.

## Why store the winner instead of computing it

The obvious objection is that for point-based endings the winner is already implied by the numbers.

**But for `submission`, `dq` and `forfeit` it's implied by nothing.** A consumer computing the winner from the scoreboard gets Bob. The whole point of the change is that it must get Carlos.

And even for point endings, deriving it means every dashboard re-implementing the tie-break ladder (points → advantages → fewer penalties → decision) and getting it subtly different. The referee's device is the only thing that was actually *at the match*; it should state the result, not leave every reader to infer it.

## Compatibility: additive, both directions already work

- **Old app, new event:** `Match.fromJson` reads named keys and ignores the rest → parses fine, shows what it showed before. No crash.
- **New app, old event:** fields absent → falls back to computing the winner from the scoreboard, exactly as today.

No kind bump. No migration.

## The referee's hands

A referee is standing over two people, one of whom just tapped, holding a phone in one hand. The common case must be **two taps**, and it must be impossible to record the wrong fighter by accident. Hold-to-finish opens an outcome sheet with **Points pre-computed and pre-selected** (one tap for a match that goes the distance) and **Submission → which fighter, in their own colours**. The clock running out still auto-finishes on points, as today.

## Four open questions I need you to decide (§8)

The sharpest one:

1. **Penalties are currently just counted.** Under IBJJF the 2nd penalty gives an advantage to the opponent, the 3rd gives two points, the 4th is a DQ. The app applies none of that. It's a pre-existing simplification and this spec doesn't change it — **but it means an automatically computed `points` winner can disagree with a real referee.** Fix the penalty rules as part of this work, or leave it and let the referee override?

Plus: is `draw` real for us? Should `dq` carry a reason? Should `ended_at` be filled in for matches that expire on the clock?

## Proposed phasing (3 PRs)

1. **Schema + model** — fields on `Match`, JSON round-trips, compatibility cases pinned as tests, tie-break ladder tested on its own. No UI, nothing changes.
2. **The outcome sheet** — hold-to-finish opens it; strings in all four locales.
3. **Showing it** — the match list says *"Carlos won by submission (armbar)"* instead of a scoreboard that lies. Update `docs/SPEC.md`, the schema other clients read.

## Summary by CodeRabbit

* **Documentation**
  * Added a revised specification for recording match outcomes.
  * Clarified winner, draw, submission, disqualification, forfeit, and completion-time handling.
  * Documented effective scoring, penalty rules, clock-expiry tie-breaking, and referee decisions.
  * Defined interaction requirements, compatibility expectations, implementation phases, and testing scope.

### DIFF
--- a/docs/specs/match-outcome.md
+++ b/docs/specs/match-outcome.md
@@ -51,7 +51,25 @@ Everything about **how the match ended** is absent: who won, and by what.
 
 ## 3. Proposal
 
-Six new fields, all optional, all only meaningful once the match is over.
+Six new fields, all only meaningful once the match is over.
+
+**Optional to *read*, required to *write*.** Every one of them is absent from
+every event published before this change, so a consumer must tolerate their
+absence forever (§7). But an event this app writes for a `finished` match must
+carry them:
+
+| Field | Required on a newly written `finished` event? |
+|---|---|
+| `method` | **Yes**, always |
+| `ended_at` | **Yes**, always |
+| `winner` | **Yes**, except when `method: "draw"` — where it must be **absent** |
+| `submission` | Optional (only meaningful with `method: "submission"`) |
+| `dq_reason` | **Yes** when `method: "dq"`, absent otherwise |
+| `dq_detail` | Optional |
+
+Without that, the app could publish a `finished` match that still says nothing
+about how it ended — which is the bug this spec exists to remove, wearing a new
+hat.
 
 ```json
 {
@@ -141,8 +159,10 @@ one.
 
 ## 4. Why store the winner rather than compute it
 
-The obvious objection: for `points`, `advantages` and `penalties`, the winner is
-already implied by the numbers. Why write it down?
+The obvious objection: for `points` and `advantages`, the winner is already
+implied by the numbers. Why write it down?
+
+(Not for `penalties` — those are no longer a way to win at all. §6.1.)
 
 **Because for `submission`, `dq` and `forfeit`, it is implied by nothing.** A
 consumer that computes the winner from the scoreboard gets Bob. The entire point
@@ -265,15 +285,22 @@ always do that: if the fighters are level on effective points **and** effective
 advantages, there is **no winner in the data**, and inventing one is exactly the
 class of lie this spec exists to remove.
 
-So:
+So, when the clock reaches zero:
 
-- **Clock expires, one fighter ahead** → finish automatically, as today, with
+- **A fighter has four or more penalties** → **never finish automatically**, even
+  if the other fighter is comfortably ahead on points. Open the sheet with `dq` /
+  `accumulated_penalties` pre-selected. Closing such a match "on points" would
+  swallow the disqualification whole: the scoreboard would name a winner and say
+  nothing about *why* the other fighter is not one. This rung comes first,
+  because §5.2 gave the disqualification to the referee, and an automatic finish
+  would take it back.
+- **Otherwise, one fighter ahead** → finish automatically, as today, with
   `method: "points"` or `"advantages"` and the winner filled in.
-- **Clock expires, fighters level** → finish the clock, keep the match on screen,
-  and **ask**: *decision (which fighter?)* or *draw*. The match is not `finished`
+- **Otherwise, fighters level** → finish the clock, keep the match on screen, and
+  **ask**: *decision (which fighter?)* or *draw*. The match is not `finished`
   until the referee says how it ended.
 
-That is one extra tap in the rarest case, and it is the only honest option: a
+That is one extra tap in the rare cases, and it is the only honest option: a
 level scoreboard is a question, not an answer.
 
 ## 7. Backwards and forwards compatibility
@@ -283,13 +310,40 @@ The fields are **additive**, and both directions already behave correctly:
 - **An old app reading a new event.** `Match.fromJson` reads named keys and
   ignores everything else, so a new event parses fine. The old app shows the
   score and "finished" — no worse than today, and no crash.
-- **A new app reading an old event.** The fields are absent → `winner` and
-  `method` are null. The app falls back to computing the winner from the
-  scoreboard, which is exactly what it does today.
+- **A new app reading an old event.** The fields are absent. The app shows the
+  match exactly as the app that published it did — see §7.1. It does **not**
+  retro-fit the new rules onto it.
 
-No version bump on the event kind. No migration. Matches published before this
-change stay readable; they simply cannot say how they ended — which is the truth
-about them.
+### 7.1 The penalty ladder is not applied retroactively
+
+This is the sharpest compatibility question in the spec, and the answer is *no*.
+
+A match published before this change has raw penalty counts and no penalty
+consequences — because the app that refereed it applied none. Reading such an
+event with §5's ladder would **rewrite a result that has already been seen**: a
+match Bob won 2–0 while Carlos carried three penalties becomes 2–2 on effective
+points, goes to advantages, and can flip to Carlos or to a draw.
+
+The argument for doing it anyway is not stupid — those matches *were* refereed
+under real IBJJF rules, where the third penalty really did concede two points, so
+the ladder arguably recovers the truth the old app failed to model. But we do not
+know what the referee did with that knowledge. They may have compensated by hand,
+in which case we would now count the same penalty twice.
+
+The costs are asymmetric. Leaving an old result showing the scoreboard it showed
+at the time is, at worst, **incomplete**. Rewriting it is **asserting something
+new about a match nobody re-refereed**. So:
+
+> **The boundary is the presence of the new fields, not a version number.**
+> An event with no `method` and no `ended_at` is a legacy event: raw scoreboard,
+> no penalty consequences, no computed winner beyond what the old app showed.
+> An event that carries them is a new event: effective scoring (§5), and a winner
+> it states outright.
+
+No `schema_version`. No bump on the event kind. No migration. The fields the app
+writes *are* the version — which is what "additive" was supposed to mean in the
+first place. Matches published before this change stay readable; they simply
+cannot say how they ended, which is the truth about them.
 
 ## 8. The referee's hands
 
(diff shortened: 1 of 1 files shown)


grunch:
Three of the four questions are decided and folded into the spec. Pushed.

## 1. Penalties become real ✅ (§5)

The IBJJF ladder is cumulative and each rung is automatic:

| Penalty | Consequence for the **opponent** |
|---|---|
| 1st | none — recorded only |
| 2nd | **+1 advantage** |
| 3rd | **+2 points** |
| 4th | **the offender is disqualified** |

This changes *who wins*, so it lands in the same phase as the outcome fields rather than being bolted on after.

**The consequences are derived, never baked in.** `f1_pen` keeps meaning "penalties given to fighter 1", and the points it concedes are computed:

```
effective_points(f1)     = 2·f1_pt2 + 3·f1_pt3 + 4·f1_pt4  +  (f2_pen ≥ 3 ? 2 : 0)
effective_advantages(f1) = f1_adv                          +  (f2_pen ≥ 2 ? 1 : 0)
```

Folding those two points into `f1_pt2` would claim fighter 1 scored a takedown they never scored — and the mistake would be **unrecoverable**, because nothing would remember where the points came from.

The **fourth penalty ends the match on its own**: `method: "dq"`, `dq_reason: "accumulated_penalties"`, winner = the other fighter.

## 2. DQ reasons — "are they standard?" is a *half* yes, and that's the whole design (§3.4)

I looked it up rather than assuming, and the answer split in two:

- **The categories are standard.** IBJJF divides severe fouls into **technical** (illegal technique, illegal grip, improper attire → disqualified from the **match**) and **disciplinary** (unsportsmanlike conduct → disqualified from the **competition**). Serious fouls instead accumulate penalties. Those categories are stable and worth encoding.
- **The infractions are not.** Whether a knee reap is illegal depends on the ruleset, the belt and the age division — and ADCC's list differs from IBJJF's again. Enumerating infractions would mean **shipping a rulebook and versioning it**.

So `dq_reason` is a three-value enum of the categories, plus `dq_detail` as free text for what actually happened:

```json
{ "method": "dq", "winner": "f1", "dq_reason": "technical_foul", "dq_detail": "knee reap" }
```

Sources: [IBJJF rulebook (PDF)](https://jjbqc.org/wp-content/uploads/2023/06/EN_IBJJF_RulesBook_MAR2022.pdf) · [rules summary](https://jiujitsulegacy.com/bjj-lifestyle/competition-tips/ibjjf-rules/)

## 3. `ended_at` — always ✅ (§3.5)

Including for a match that runs out the clock. Computable there, but writing it anyway means every finished match has the same shape and no consumer has to special-case one.

## Still open: is `draw` real for us?

You didn't say, so I've **kept it** — it costs nothing, and some rulesets (and friendly in-house comps) use it. IBJJF itself has no draws in most brackets; a level match goes to referee decision. Say the word and it goes.

Once you're happy with this, phase 1 is the schema + model + the penalty ladder + the tie-break ladder, all tested, with no UI.

grunch:
Decided and pushed. **All four questions are closed** — the spec is ready to implement.

## Draws are real, and a level match is the referees' to call

- More effective **points** wins.
- Level on points → more effective **advantages** wins.
- Level on both → **the referees decide**: they may name a winner (`method: "decision"`) or call it even (`method: "draw"`, no winner).

## Encoding your rule surfaced a real bug in my original ladder

I had a third rung: *"fewer penalties wins"*. **That's wrong now**, and your simpler ladder is right precisely because of it.

§5 turns penalties **into** advantages and points — a 2nd penalty *is* an advantage for the opponent, a 3rd *is* two points. They're already **inside** the numbers being compared at the first two rungs. Using the raw penalty count again would **count the same penalty twice**, and it would silently *reverse* matches: a fighter whose third penalty already handed their opponent two points could then lose a **second** time for having the higher count.

So `penalties` is gone as a `method`. It cannot be the reason anybody won.

## And a consequence worth flagging: the app can no longer always finish a match by itself

If the clock expires with the fighters **level**, there is **no winner in the data** — and inventing one is exactly the class of lie this spec exists to remove.

So: clock expires with someone ahead → finishes automatically, as today. Clock expires level → the clock stops, the match stays on screen, and it **asks**: *decision (which fighter?)* or *draw*. The match isn't `finished` until the referee says how it ended.

One extra tap, in the rarest case. **A level scoreboard is a question, not an answer.**

---

Say the word and I'll start phase 1: schema + model + penalty ladder + tie-break ladder, fully tested, no UI.

grunch:
All three decided and pushed. The spec is complete.

## 1. The fourth penalty records, it does not decide (§5.2)

**Your call is better than my proposal, and for the reason that matters:** the app should not *infer* a disqualification. The referee is the authority.

It also disarms the mis-tap. An auto-disqualifying fourth penalty would make that button the one control on the mat that **ends a match and decides who loses** — and a fat thumb would disqualify a fighter, publish it to Nostr instantly, and be **unrecoverable** (a finished match is read-only, and `canUndo` switches off the moment a match stops running).

So it goes on the scoreboard like the first three:

- a mis-tap is undone by **the hold that already exists**, because the match is still running;
- the referee ends the match deliberately, and the outcome sheet **pre-selects** `dq` / `accumulated_penalties` with the other fighter as winner — one tap to confirm what the scoreboard already says;
- **four penalties are right there in the event**, so a dashboard can render "disqualified" without needing anything from us.

## 2. Draws encoded once (§3.1)

`winner` is `f1`, `f2`, or **absent**. A draw is `method: "draw"` with no winner.

Keeping `"draw"` as a third `winner` value would let an event say `winner: "f1"` **and** `method: "draw"` at the same time — and every consumer would resolve that contradiction differently.

## 3. The referee sees the effective score (§5.1)

When Carlos's third penalty gives Bob two points, **Bob's number goes 4 → 6 on screen** — that jump *is* the confirmation the penalty landed. Carlos's penalty badge still shows the raw count, because that's what he was actually given. Holding the penalty button back **removes those two points again**: they were never Bob's, they were Carlos's penalty.

## One thing I want on the record before we build (§10.1)

**A finished match is read-only today.** `_buildReadOnlyFooter`; `finishMatch` and `cancelMatch` are no-ops once `isFinished`. So a match the clock closed with the wrong winner — a penalty entered against the wrong fighter, say — **stays wrong forever**.

I removed the "the referee can amend it" line I'd written, because the app can't. The mechanics are cheap (addressable event, republish, monotonic `created_at` already handles it); the cost is UI. **It goes in phase 3, and until it exists the app should not pretend.**

---

Ready for phase 1 whenever you are: schema + model + penalty ladder + tie-break ladder, fully tested, no UI.

grunch:
All six verified against the spec. **Five were mine to begin with**; one needed a decision, which we took: **the penalty ladder is not applied retroactively** (§7.1).

## The one that mattered

A new app reading an old event was going to compute the winner with §5's ladder. That **rewrites results people have already seen**: a match Bob won 2–0 while Carlos carried three penalties becomes 2–2 on effective points, goes to advantages, and can flip to Carlos — or to a draw.

Those matches *were* refereed under real IBJJF rules, where the third penalty really did concede two points, so the ladder arguably recovers a truth the old app failed to model. **But we don't know what the referee did with that knowledge** — they may have compensated by hand, and we'd now count the same penalty twice.

The costs are asymmetric. Leaving an old result showing the scoreboard it showed at the time is, at worst, **incomplete**. Rewriting it is **asserting something new about a match nobody re-refereed**.

> **The boundary is the presence of the new fields, not a version number.**
> No `method` and no `ended_at` → legacy event: raw scoreboard, no penalty consequences.
> Carries them → new event: effective scoring, and a winner it states outright.

No `schema_version`, no kind bump, no migration. **The fields the app writes *are* the version** — which is what "additive" was supposed to mean in the first place.

## The other five

- **Clock expiry vs. four penalties** was genuinely unspecified. Now: with four or more penalties the clock **never** finishes the match automatically, however far ahead the other fighter is. Closing it "on points" would swallow the disqualification whole — naming a winner and saying nothing about why the other fighter isn't one. §5.2 gave that call to the referee; an automatic finish would take it back.
- **"Scoring, penalties, advantages: untouched"** was **false** the moment penalties gained consequences. Split it honestly: the raw record and the referee's inputs are untouched; what those numbers *add up to* is not, and that's the point of the spec rather than an accident of it.
- **`penalties` was still listed as a derivable way to win** in §4 — we removed it as a method precisely because §5 already folds penalties into the points and advantages being compared.
- **The new fields are optional to *read*, required to *write*.** A `finished` event this app publishes must carry `method` and `ended_at`, and `winner` unless it's a draw. Otherwise the app could publish a finished match that still says nothing about how it ended — the original bug, wearing a new hat.
- One stale `§6` → `§7` cross-reference.

PR #91, opened 2026-07-14
feat(match): Phase 1 — record how a match was won, and make penalties count

grunch:
Phase 1 of [the outcome spec](docs/specs/match-outcome.md): **the model, and nothing else.** No UI sets these fields yet, so nothing the app does changes. This is the phase that decides *who wins*, so it's the one that has to be right.

## What's here

**Outcomes.** `Match` now carries `winner`, `method`, `submission`, `dqReason`, `dqDetail`, `endedAt` — so an event can finally say that Carlos submitted Bob, instead of publishing a 4–0 scoreboard naming Bob.

**Penalties become real.** The IBJJF ladder is cumulative and the app applied none of it: 2nd penalty → an advantage to the opponent, 3rd → two points, 4th → a disqualification.

The consequences are **derived, never baked in**. `f1Pt2` still means *"takedowns fighter 1 scored"*; the conceded points live in `f1EffectivePoints`. Folding them into the counters would claim a takedown that never happened — and the mistake would be **unrecoverable**, because nothing would remember where the points came from.

**The fourth penalty is recognised, not acted on** (`hasDisqualifyingPenalties`). A penalty that ended a match by itself would make that button the one control on the mat that decides who loses, and a mis-tap would be unrecoverable.

**`scoreboardWinner` refuses to name a winner when the fighters are level** — and that `null` is the entire point. A level match has **no winner in the data**, and inventing one is the lie this model exists to prevent.

**Penalties are never a tiebreak of their own.** They're already *inside* the points and advantages being compared, so counting them again would count the same penalty twice — a fighter whose third penalty had already handed two points away could then lose *again* for having the higher count.

## Legacy matches are not re-refereed

A match finished before outcomes existed keeps its old arithmetic. Applying the ladder retroactively would **rewrite results people have already seen**: a 2–0 win with three penalties against the loser becomes 4–4, goes to advantages, and flips.

## Implementing it turned up a hole in the spec's own rule

The spec said: *"no `method` and no `ended_at` → legacy event"*.

**That describes the app's own live match too.** Those fields are only filled in when a match *ends* — so taken literally, the rule would have hidden the penalty points from the referee **precisely while they were refereeing**, which is the one moment they must see them.

`isLegacyResult` is therefore *"finished or canceled, **and** silent about how"*. A match in progress has no result to preserve; it's being refereed **now**, by this app, under these rules. The spec is corrected in the same commit.

## Test plan

- [x] `flutter test` — **172 passing** (26 new)
- [x] The penalty ladder, rung by rung, including the 4th adding nothing arithmetically
- [x] A penalty **flipping the match** to the fighter who was behind
- [x] The tie-break refusing to name a winner when level
- [x] Legacy events keeping their old arithmetic; a **live** match never treated as legacy
- [x] Every inconsistent outcome rejected: a draw with a winner, a method without one, a `dq` without a reason, an outcome without an `ended_at`
- [x] JSON round-trips; absent keys (not nulls) on an unfinished match; an unknown `method` is a hard error rather than a silent null

## What's next

Phase 2 (the outcome sheet) depends on this, so it branches from here.

## Summary by CodeRabbit

* **New Features**
  * Added explicit match outcomes, including winner, decision method, submissions, disqualification details, and end time.
  * Added support for accurately applying penalty effects to completed match results.
  * Preserved legacy scoring behavior for completed or canceled matches without recorded outcomes.
  * Live matches without an outcome remain subject to current penalty effects.

* **Documentation**
  * Clarified the definition and handling of legacy match events.

### DIFF
--- a/docs/specs/match-outcome.md
+++ b/docs/specs/match-outcome.md
@@ -335,11 +335,19 @@ at the time is, at worst, **incomplete**. Rewriting it is **asserting something
 new about a match nobody re-refereed**. So:
 
 > **The boundary is the presence of the new fields, not a version number.**
-> A **finished or canceled** match with no `method` is a legacy event: raw
+> A **finished or canceled** match with no `ended_at` is a legacy event: raw
 > scoreboard, no penalty consequences, no computed winner beyond what the old
-> app showed. Anything else — a match still in progress, or a finished one that
-> states its method — uses effective scoring (§5) and, where it has one, the
-> winner it states outright.
+> app showed. Anything else — a match still in progress, or one this app ended —
+> uses effective scoring (§5) and, where it has one, the winner it states
+> outright.
+
+The test is **`ended_at`**, not `method`. A legacy event has neither, but a
+match *this* app ends always stamps `ended_at` — even in the one case where it
+cannot yet name a method, because the fighters are level and that is the
+referees' to call (§6.2). Keying the rule off `method` would make the app
+contradict itself the instant a referee pressed Finish: a 0–0 match against
+three penalties reads 2–0 while it runs, and would snap back to 0–0 the moment
+it stopped.
 
 **A live match is never legacy.** Taken literally, "no `method` and no
 `ended_at`" would describe the app's *own* match while it is being refereed:
--- a/docs/specs/match-outcome.md
+++ b/docs/specs/match-outcome.md
@@ -335,10 +335,19 @@ at the time is, at worst, **incomplete**. Rewriting it is **asserting something
 new about a match nobody re-refereed**. So:
 
 > **The boundary is the presence of the new fields, not a version number.**
-> An event with no `method` and no `ended_at` is a legacy event: raw scoreboard,
-> no penalty consequences, no computed winner beyond what the old app showed.
-> An event that carries them is a new event: effective scoring (§5), and a winner
-> it states outright.
+> A **finished or canceled** match with no `method` is a legacy event: raw
+> scoreboard, no penalty consequences, no computed winner beyond what the old
+> app showed. Anything else — a match still in progress, or a finished one that
+> states its method — uses effective scoring (§5) and, where it has one, the
+> winner it states outright.
+
+**A live match is never legacy.** Taken literally, "no `method` and no
+`ended_at`" would describe the app's *own* match while it is being refereed:
+those fields are only filled in when it ends. Reading that as legacy would hide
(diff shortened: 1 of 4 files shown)


grunch:
All three findings verified against the code. **Two were real and are fixed; one I'm skipping with a reason.**

## 1. A newly finished match was classified as legacy (codex, P1) — real, and the sharpest one

Confirmed. The legacy boundary keyed off `method`, and the Finish button doesn't set one. So **the instant a referee pressed Finish, the match became indistinguishable from an event published before outcomes existed** — and the penalty points they had just watched land vanished.

Concretely: a 0–0 match against three penalties reads **2–0** while it runs, and snapped back to **0–0** the moment it stopped. The app contradicted itself, and a consumer would compute a different winner than the referee saw.

**The test is `endedAt` now.** A legacy event has neither field; a match *this* app ends always stamps `endedAt` — even in the one case where it cannot yet name a method, because the fighters are level and that's the referees' to call. Finishing also records what the scoreboard already knows, instead of stopping silently.

The spec's own §7.1 said "no `method` and no `ended_at`", which was ambiguous enough to permit the bug. It now says `ended_at`, and says why.

## 2. Validation holes (CodeRabbit) — real

- **A `winner` with no `method` passed.** It says who won and not what they won by, which is half an answer: a consumer cannot tell a submission from a points decision. Now refused.
- **`endedAt` was never checked against `startAt`**, unlike the `pausedAt` check right below it — so a match could end before it began. Now refused.

That second guard immediately **caught my own phase-3 test fixtures**, which pinned a fixed end time onto a match starting *now* — a match that ended twenty-six years before it began. The guard was right; the tests were wrong.

**Three regression tests, all red without the fix.**

## 3. Effective scoring not surfaced in the score panels (codex, P2) — skipping, with a reason

Correct as an observation, but it's **out of scope for this PR by design**. Phase 1 is explicitly model-only: *"No UI: nothing sets the new fields yet, so nothing changes."*

The UI work is done and already open:

- **#92** puts the effective score on the match screen (the referee sees a penalty become the opponent's points, and that jump is how they know it landed).
- **#93** puts it on the home card — and fixes the deeper version of this bug, where the card highlighted **the bigger number** and therefore lit up the fighter who *lost* a submission.

Merging them together closes the gap completely; doing it here would just move Phase 2's diff into Phase 1's.

---

Green across the chain: **175** (#91) / 191 (#92) / **194** (#93). The fix is propagated through all three branches.

PR #92, opened 2026-07-14
feat(match): Phase 2 — ask the referee how the match ended

grunch:
Phase 2 of [the outcome spec](docs/specs/match-outcome.md). **Depends on #91** (Phase 1) — it targets that branch.

**A match can no longer be finished without saying how.** `finishMatch()` is gone; `finishWith(MatchOutcome)` took its place. That's the point: a match that stops with only a status publishes a scoreboard that can name the wrong fighter.

## The sheet

Holding *Finish* opens it instead of ending the match.

- **The scoreboard's own verdict is pre-selected** — a match that goes the distance is **one tap**.
- **Submission comes first**, because it's what the sport is for, and the reason a scoreboard on its own can lie. It asks *which fighter* **in their own colours** — the same two the referee has been tapping all match, never by position.
- **The free text is always skippable.** Nobody should be blocked from ending a match because the app has never heard of a *baratoplata*.

## The clock no longer finishes a match it cannot decide

- **Fighters level** → no winner exists *in the data*, and inventing one is the lie this whole feature exists to remove. The clock stops, the match stays open, the sheet asks.
- **A fighter on four penalties** → closing that "on points" would **swallow the disqualification whole**: it would name a winner and say nothing about why the other fighter isn't one. Only a referee may call a DQ, so the app offers it and waits.
- **Otherwise** → finishes automatically, as today, with the winner and method filled in.

**Dismissing the sheet leaves the match open** — the honest state of a match nobody has decided, and it lets the referee go back and fix the score before committing to a result.

## The referee now sees the effective score

A penalty against the opponent has already become points, and **that jump is how the referee knows it landed**. The penalty badge still shows the **raw** count, because that's what the fighter was actually given.

## Test plan

- [x] `flutter test` — **188 passing** (19 new)
- [x] Holding Finish asks instead of ending
- [x] **A submission ends the match against the scoreboard** — Pana leads 4–0 and Buchecha wins
- [x] The scoreboard result is one tap
- [x] Dismissing the sheet leaves the match open
- [x] Level clock expiry → asks; leader → finishes automatically
- [x] Four penalties → clock refuses to finish on points, and the sheet offers the DQ
- [x] **A mis-tapped fourth penalty is taken back by holding** — the reason it doesn't auto-DQ
- [x] The scoreboard shows the two points a third penalty conceded
- [x] Strings in `en`, `es`, `pt`, `ja`


grunch:
All three findings verified. **All three were real, and all three are fixed** — each with a regression test that goes red without it.

## 1. The sheet never appeared for a match whose clock expired while the app was closed

Confirmed, and this one is a dead end for the referee. The notifier settles the expiry **in its constructor** — before the widget exists — so the screen's first watched state already has `awaitsOutcome == true`. `ref.listen` only reports *changes*, so it never fired.

The match would sit at **00:00, in progress, forever**, until someone thought to hold Finish. It now asks on the first frame.

## 2. Both fighters at four penalties → an arbitrary winner

Confirmed, and reachable **precisely because** the fourth penalty leaves the match running (which is the design we chose deliberately). The suggestion assumed fighter 1 was the offender and offered a one-tap result in which **fighter 2 wins by disqualification**.

An arbitrary winner is the worst kind, because **it looks decided** — a hurried referee taps the pre-selected option and publishes a result nobody chose. With both disqualified the scoreboard names nobody, so the app now offers **nothing** and the referee has to choose.

## 3. A match the clock ended was stamped with the moment the app noticed

Confirmed. Reopen a match whose five minutes ran out **yesterday**, and `ended_at` recorded **today** — and every consumer would have believed it. The spec is explicit that `ended_at` is not derivable, which cuts both ways: it also has to be *right*.

The clock path now stamps the real expiry (`startAt + duration`, with `startAt` already moved forward by any stoppage). A referee ending a match in front of them still stamps `now`, which is what that means.

## While propagating to #93

The merge left `finishWith` taking an `endedAt` it never passed on to `_setOutcome` (Phase 3 had refactored the body out from under it), and my own conflict resolution ate a closing brace in the screen tests. Both caught by the analyzer before they reached anyone.

---

Green across the chain: 175 (#91) / **196** (#92) / **199** (#93).

PR #93, opened 2026-07-14
feat(match): Phase 3 — show the result, and let a wrong one be corrected

grunch:
Phase 3 of [the outcome spec](docs/specs/match-outcome.md) — the last one. **Depends on #92 → #91**, so the chain merges 91 → 92 → this.

## The home card was lighting up the fighter who lost

This is the original bug in its purest form. The card highlighted **the bigger number**:

```dart
color: match.f1Score > match.f2Score ? tk.accent : tk.muted,
```

A match that ended on a submission shows **the loser's numbers as the bigger ones** — so the card lit up the fighter who lost, in the winner's colour.

It now names the winner **the event names**, and only falls back to the scoreboard while a match is genuinely undecided. The numbers themselves are the **effective** ones, so a penalty that conceded points shows up there too.

## Both surfaces now say *how* it ended

*"Buchecha · Submission (armbar)"* — on the card and in the match screen's footer. A score row cannot say that, which is the entire premise of the feature.

## A wrong result can be corrected (§10.1)

This was the gap I flagged in the spec and refused to promise away: a result the clock closed against the wrong fighter — a penalty entered against the wrong man, say — **would stay wrong forever**. A finished match was read-only, and the event was already out on the relays.

**Amend result** reopens the outcome sheet and republishes. The mechanics are cheap, as predicted: the event is addressable, so the correction **supersedes the mistake on every relay**, and the monotonic `created_at` makes sure it wins.

A **canceled** match has no amend button — it isn't a result, it's the absence of one.

## `docs/SPEC.md` is updated

The schema other clients read now documents the outcome fields, the penalty ladder, the effective-score formulas, and the rule that **legacy events are not re-refereed**.

## Test plan

- [x] `flutter test` — **191 passing**
- [x] A finished match says how it ended, not just that it did
- [x] **A wrong result can be corrected**, and the correction republishes
- [x] A canceled match has nothing to amend
- [x] Android APK builds

## The feature, end to end

Bob leads 4–0. Carlos armbars him. The referee holds Finish, taps **Submission**, taps **Carlos** — and every relay, every dashboard, and the app's own card now say **Carlos won by submission (armbar)**, with Bob's 4–0 still on the record where it belongs.

## Summary by CodeRabbit

* **New Features**
  * Match cards now show how each match ended, including winner, method, and submission details.
  * Finished match results can be amended and republished.
  * Finished and canceled match views now clearly distinguish available actions.
  * Score and advantage displays reflect effective values, including penalty adjustments.
  * Added localized outcome labels in English, Spanish, Japanese, and Portuguese.

* **Documentation**
  * Updated scoring, outcome, penalty, and result-amendment guidance.

* **Tests**
  * Added coverage for outcome display, result amendments, and canceled-match behavior.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -45,11 +45,56 @@ Fetch events from nostr relays in real time, the events we are fetching are auth
   "f1_adv": 0,
   "f2_adv": 0,
   "f1_pen": 0,
-  "f2_pen": 0
+  "f2_pen": 0,
+
+  "winner": "f1",
+  "method": "submission",
+  "submission": "armbar",
+  "dq_reason": null,
+  "dq_detail": null,
+  "ended_at": 1700000180
 }
 ```
 
-> **Score = pt2×2 + pt3×3 + pt4×4** (per fighter)
+> **Raw score = pt2×2 + pt3×3 + pt4×4** (per fighter)
+>
+> **Effective score = raw score + 2 if the *opponent* has 3+ penalties**
+> **Effective advantages = adv + 1 if the *opponent* has 2+ penalties**
+
+### Outcome (how the match was won)
+
+The scoreboard cannot name the winner on its own: a fighter can lead 4–0 and
+lose to an armbar. These fields say what actually happened, and they are
+**absent** from events published before they existed — a consumer must tolerate
+that forever.
+
+| Field | Values |
+|---|---|
+| `winner` | `f1` \| `f2`. **Absent** while unfinished, when canceled, and on a draw. |
+| `method` | `submission` \| `points` \| `advantages` \| `decision` \| `dq` \| `forfeit` \| `draw` |
+| `submission` | Free text (`"armbar"`). Optional, and only with `method: submission`. |
+| `dq_reason` | `accumulated_penalties` \| `technical_foul` \| `disciplinary_foul`. Required with `method: dq`. |
+| `dq_detail` | Free text (`"knee reap"`). Optional. |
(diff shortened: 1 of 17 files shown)


grunch:
@coderabbitai review

grunch:
@coderabbitai review

grunch:
All three comments verified — they're **two distinct bugs** (codex and CodeRabbit both flagged the `endedAt` one). **Both real, both fixed**, each with a regression test that goes red without it.

## 1. A correction is not an ending

`amendOutcome` overwrote `endedAt` with the moment of the edit. Correct the winner of **yesterday's** match **today**, and every consumer would believe the match ended today.

The Nostr event's own `created_at` already records when the correction went out. `ended_at` is when the **fight** stopped. It's preserved now — and a legacy result, which has none to preserve, gets one: the best answer available, and better than none.

## 2. The referee's answer could vanish silently

The sheet decided between *finish* and *amend* from the state it was **opened** with. If the clock ran out while it was open, the notifier finished the match on points behind it — and `finishWith` then became a **silent no-op**. The referee's chosen submission simply disappeared, leaving the fighter the scoreboard favoured published as the winner.

That's the original bug wearing a disguise: the scoreboard wins, quietly, over what actually happened. The state is re-read after the sheet closes.

## A note on the second test, because it matters

My first version of that regression test was **worthless**. It tried to expire the clock inside the widget test — but the provider reads the **real wall clock**, which a widget test cannot advance. So it passed with *and* without the fix.

I only caught it because I check that a regression test goes **red** before trusting it. It now drives the match to `finished` behind the open sheet, which is the condition that actually breaks, and it *does* go red without the fix.

---

**199 tests** green; analyzer unchanged.

PR #98, opened 2026-07-14
fix: resume must never strand the relay transport

grunch:
Fixes the "app has no connection until I kill it" bug.

## The symptom

Sometimes, after returning to the app, creating a match fails with a no-connection
error. Retry fails too — every time. Killing and reopening the app fixes it
instantly. Wi-Fi is fine and every other app works.

## The cause

Our own resume hook, racing a bug in nostr-sdk 0.44.

On every `AppLifecycleState.resumed`, the app calls `reconnectAll()` —
`disconnect()` + `connect()` on the pool. In nostr-sdk 0.44 that revival can
lose a race against its own teardown:

1. `disconnect()` stores a termination permit (a tokio `Notify`) and returns
   with the relay already `Terminated`.
2. `connect()` flips it to `Pending` and spawns a connection task — which sees
   the **old task still registered**, and declines.
3. The old task wakes, reads `Pending` (not terminated → it keeps going), marks
   the relay `Disconnected`, and dies in its backoff sleep by consuming the
   stored permit.

End state: **no task, status `Disconnected`** — and `connect()` is a no-op on
`Disconnected`. Nothing ever dials again. Dart's `connectedRelays` stays empty,
so every publish throws before touching the network. Only a process restart
rebuilds the world. It is a race, hence "sometimes".

## Reproduced before fixed

The new drill (`resume never strands the transport`) hammers `reconnectAll()`
twenty times against a live relay — exactly as the lifecycle hook fires it —
and asserts the transport comes back and a score lands.

- **Red** on the old code: `connectedRelays` stays empty for 20s+, the referee's
  Retry against a dead app.
- A first fix attempt (re-nudging stragglers with more disconnect/connect
  pairs, 20 rounds) **stayed red**: each nudge re-enters the same race. Worth
  recording so nobody retries that road.
- A probe confirmed the strand is Rust-side (`relay_connected()` empty), not a
  lost Dart notification.

## The fix

The one sequence with no revival in it: **remove every relay and add it back.**
A fresh `Relay` has a fresh `Notify` (no stored permit), no zombie task, and a
new status watcher. Pool subscriptions survive by design — nostr-relay-pool
makes newly added relays inherit them.

Also hardened all three notification loops against broadcast `Lagged`: they
used `while let Ok`, so falling behind by one channel capacity killed the loop
**silently** — the status-watcher variant of that is a Dart-side connection
view frozen until restart, the same symptom by another road. They now skip what
they missed (the watcher re-reports the relay's current status) and exit only
when the channel closes.

## Tests

- New drill red→green; the four #78 convergence drills still pass (44 rust-tagged).
- Full suite: 227 passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relay reconnect behavior so repeated reconnects and app resume no longer strand transports or resurrect removed relays.
  * Updated relay status and event streaming to recover cleanly after missed/broadcast-lagged updates and to continue emitting fresh status snapshots.
  * Enhanced relays’ reported connection state after temporary delays to keep the client view accurate.
* **Tests**
  * Added an integration convergence drill covering repeated resume/reconnect timing races, verifying the relay reconnects and can receive events afterward.

### DIFF
--- a/rust/src/api/relay.rs
+++ b/rust/src/api/relay.rs
@@ -53,6 +53,19 @@ pub struct RelayStatusData {
     pub connected: bool,
 }
 
+/// Every URL the caller has asked for, whether or not the pool holds it.
+///
+/// The pool is not this list: a relay whose re-add failed mid-reconnect is
+/// gone from the pool, and a reconnect that rebuilds "whatever the pool has"
+/// would never try it again — the relay would be silently gone until the
+/// process restarts. This registry is what was *asked for*; the pool is what
+/// *succeeded*. Reconnect rebuilds from the former.
+static CONFIGURED: OnceLock<std::sync::Mutex<std::collections::BTreeSet<String>>> = OnceLock::new();
+
+fn configured() -> &'static std::sync::Mutex<std::collections::BTreeSet<String>> {
+    CONFIGURED.get_or_init(|| std::sync::Mutex::new(std::collections::BTreeSet::new()))
+}
+
 /// A NIP-01 `REQ` filter. Empty vectors and `None` mean "unconstrained" — the
 /// same shape the Dart `Filter` has, so the mapping stays honest.
 pub struct FilterData {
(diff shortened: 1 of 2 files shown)


grunch:
@coderabbitai review

PR #109, opened 2026-07-17
ci: also build a signed .aab for Google Play in the release workflow

grunch:
## What

Extends `release.yml` so each `v*` tag also produces a **signed Android App Bundle (`.aab`)** for Google Play, alongside the existing sideload APK.

## Why

Google Play only accepts App Bundles, not APKs. The release workflow previously built just `choke-<tag>.apk` (arm64, for direct download), so there was no artifact you could upload to the Play Console.

## Changes

- **Build the AAB** with `flutter build appbundle --release`, signed with the same upload key as the APK (via `android/key.properties`, already set up in the workflow).
- **All ABIs on purpose:** unlike the arm64-only sideload APK, the bundle keeps every ABI (the default). Play generates optimized per-device APKs from the bundle, so each user still downloads only their own architecture — including all ABIs costs nothing at install time and covers arm64, armeabi-v7a, and x86_64.
- **Free disk space step:** building the all-ABI bundle plus the APK, NDK, and caches would otherwise exhaust the runner (`No space left on device`). Mirrors the existing step in the `android` job of `rust.yml`.
- **Attach both artifacts** (`.apk` + `.aab`) to the GitHub release.

## Test plan

- [ ] Push a `v*` tag and confirm the workflow builds both `choke-<tag>.apk` and `choke-<tag>.aab`.
- [ ] Confirm the `.aab` is signed (`jarsigner -verify` / `bundletool`) with the upload key.
- [ ] Upload the `.aab` to the Play Console internal testing track and confirm it is accepted.
- [ ] Confirm the runner does not run out of disk during the build.

## Summary by CodeRabbit

* **New Features**
  * Releases now include Android App Bundle (AAB) artifacts alongside APK files.
  * AAB files are published with versioned filenames for easier identification.
  * Both APK and AAB are uploaded as part of the same release.
* **Bug Fixes**
  * Hardened release/tag handling with stricter semantic version validation and safer output generation.
  * Improved changelog generation and release-step reliability.
  * Added automated runner disk cleanup to prevent build failures due to low disk space.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,6 +149,19 @@ jobs:
           git commit -m "chore: update changelog and version for ${{ steps.version.outputs.full_tag }}"
           git push origin main
 
+      # The App Bundle packs native code for every ABI, so Cargokit
+      # cross-compiles the Rust crate three times (arm, arm64, x86_64). Together
+      # with the APK build, the NDK, and the Gradle/cargo caches this exhausts
+      # the ~14 GB free on the runner ("No space left on device"). Reclaim space
+      # from preinstalled toolchains this job never uses.
+      - name: Free disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/.ghcup \
+            /usr/share/swift /usr/local/share/powershell \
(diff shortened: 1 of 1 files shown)


grunch:
@codex review

PR #112, opened 2026-07-18
feat(account): share a bjjscore.live live-board link

grunch:
## Summary

Adds a **"Share live board"** button on the Account screen (under the npub, next to Copy / Show QR) that opens the **native share sheet** with a link to the public board carrying this organizer's key:

```
https://bjjscore.live/?npub=<npub>
```

Opening that link drops a spectator straight onto the user's matches — **no key to copy or paste** on the web. This is step 2 of the "share my matches" flow; the receiving half already shipped in choke-scoreboard (`readSharedPubkey` / `buildShareLink`), and the URL format here matches it, kept in one place via `kLiveBoardBaseUrl` / `liveBoardShareUrl()`.

## Changes

- **UI:** filled full-width share CTA below Copy/QR in `account_screen.dart`; calls the platform share sheet with a localized message + the link, passing a `sharePositionOrigin` so the iPad popover has an anchor.
- **Dependency:** `share_plus: ^10.1.4` (+ its Windows plugin registration, regenerated by `pub get`).
- **l10n:** `shareLiveBoard` and `shareLiveBoardMessage` added to `en/es/pt/ja` and regenerated.

## Test plan
- [x] `flutter gen-l10n` — getters generated
- [x] `flutter analyze` — no new errors/warnings from these changes (pre-existing infos/warnings live in other files)
- [ ] Manual: on a device, tap **Share live board** → the share sheet opens with the `?npub=` link; opening it on bjjscore.live auto-loads this organizer's matches

_Note: verified via gen-l10n + static analysis; the share sheet itself needs a real device/emulator (none in this environment)._

## Summary by CodeRabbit

* **New Features**
  * Added a “Share live board” action to the account screen.
  * Users can share a live BJJ scoreboard link via the device’s system share sheet.
* **Localization**
  * Added translated share labels and messages (English, Spanish, Japanese, Portuguese), plus a localized error message for share failures.
* **Platform Support**
  * Enabled sharing support on Windows alongside other supported platforms.

### DIFF
--- a/lib/features/account/account_screen.dart
+++ b/lib/features/account/account_screen.dart
@@ -2,9 +2,19 @@ import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter/services.dart';
 import 'package:qr_flutter/qr_flutter.dart';
+import 'package:share_plus/share_plus.dart';
 import 'package:choke/l10n/generated/app_localizations.dart';
 import '../../shared/theme/app_theme.dart';
 import '../../services/key_management/key_manager.dart';
+
+/// Base of the public live board. A shared link carries the organizer's npub in
+/// the query string (`?npub=…`), so opening it drops the spectator straight onto
+/// this user's matches with nothing to paste. Must match the reader in
+/// choke-scoreboard (`buildShareLink` / `readSharedPubkey`).
+const String kLiveBoardBaseUrl = 'https://bjjscore.live';
+
+/// Build the share link for an organizer's npub.
+String liveBoardShareUrl(String npub) => '$kLiveBoardBaseUrl/?npub=$npub';
 
 class AccountScreen extends ConsumerStatefulWidget {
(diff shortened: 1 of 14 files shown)


grunch:
@coderabbitai review

grunch:
Fixed in 9b8613d.

The share handler now wraps `Share.share(...)` in a `try/catch`: on a `PlatformException` (share sheet fails to open — no handler, transient platform error) it no longer escapes the tap callback as an unhandled async error. Instead it:

- logs only the exception **code** via `debugPrint` (never the message, which could echo shared content), and
- shows a localized error snackbar (`shareFailed`) using the screen's existing `ScaffoldMessenger` + `colorScheme.error` conventions, guarded by a `mounted` check after the await.

The success path (message, subject, `sharePositionOrigin`) is unchanged. Added the `shareFailed` string in en/es/pt/ja and regenerated localizations. `flutter analyze` on the file is clean (only pre-existing `withOpacity` infos remain).

PR #119, opened 2026-07-18
fix: don't let a slow or failing relay delay scoreboard updates

grunch:
## Symptom

Advantage/penalty presses *sometimes* took seconds to reach the live board, while point presses felt instant.

## Root cause

All buttons share one code path (`_score` → `_publishState`), so the difference was **timing, not the button**: adv/pen presses usually follow another press (score → advantage), so they landed while a publish was still in flight and queued behind it. Two compounding causes:

1. **`publishEvent` awaited every relay's verdict.** `Future.wait` over all connected relays, even though its contract says one ack = success. One slow relay gated the serialized publish queue for its full round-trip.
2. **The failure path held fresh actions hostage to backoff.** If the in-flight publish failed, `_drainOutbox` scheduled a backoff and broke — even when a *fresh* action had landed mid-flight (its retry-reset in `_publishState` was a no-op during the await). The new state then waited out a delay meant for the stale one.

## Fix

- `publishEvent` now resolves on the **first ack**. Stragglers keep publishing in the background; their acks still land in `_markAccepted` and the existing resend sweep converges any relay that missed the event. Zero acceptances still throws.
- `_drainOutbox`'s catch now checks whether the outbox changed mid-flight: a fresh state publishes immediately; only an unchanged state waits out the backoff.

## Tests

- New regression test: `publishEvent` must resolve promptly when one relay acks and another never answers (with the old code it hangs and times out).
- Full provider + service suites pass (53 tests).

## Test plan

- [ ] CI green (crate, bridge, android)
- [ ] Manual: on a device with one slow relay configured, hammer score → advantage in quick succession; the board should update per press with no multi-second stalls.

## Summary by CodeRabbit

- **Bug Fixes**
  - Event publishing now completes as soon as one relay confirms delivery, instead of waiting on slower relays.
  - Unresponsive relays are automatically timed out, preventing publishing from hanging indefinitely.
  - Newer match updates are prioritized when an earlier delivery attempt fails.
  - Relay connections are refreshed after repeated publishing failures.

- **New Features**
  - Added configurable publishing timeout support, defaulting to five seconds.

- **Tests**
  - Added coverage for fast acknowledgments, stuck relays, and publishing timeouts.

### DIFF
--- a/lib/features/match/providers/match_control_provider.dart
+++ b/lib/features/match/providers/match_control_provider.dart
@@ -474,6 +474,11 @@ class MatchControlNotifier extends StateNotifier<MatchControlState> {
         debugPrint(
             'MatchControl: publish failed (attempt ${_retryAttempt + 1}): $e');
         _retryAttempt++;
+        // A fresh action landed while this send was in flight: its retry-reset
+        // in _publishState was a no-op (we were mid-await), so backing off here
+        // would hold the NEW state hostage to a delay meant for the stale one.
+        // Send it now; only an unchanged outbox waits out the backoff.
+        if (!identical(_outbox, match)) continue;
         final delay = _retryBaseDelay * (1 << (_retryAttempt - 1).clamp(0, 4));
         _retryTimer = Timer(
           delay > _retryMaxDelay ? _retryMaxDelay : delay,
--- a/lib/features/match/providers/match_control_provider.dart
+++ b/lib/features/match/providers/match_control_provider.dart
@@ -479,6 +479,12 @@ class MatchControlNotifier extends StateNotifier<MatchControlState> {
         // would hold the NEW state hostage to a delay meant for the stale one.
         // Send it now; only an unchanged outbox waits out the backoff.
         if (!identical(_outbox, match)) continue;
+        // Two failures in a row usually mean sockets that look open but are
+        // dead (dropped silently while the mat was quiet). Recycle them; the
+        // reconnect listener drains the outbox as soon as a relay is back.
+        if (_retryAttempt >= 2) {
+          unawaited(_nostrService.reconnectAll());
+        }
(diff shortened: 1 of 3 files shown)


grunch:
@coderabbitai review

grunch:
@coderabbitai review

grunch:
Fixed in 7674d03.

**@ermeme — resend path could still wait forever (blocker).** `_resendPendingTo()` now applies the same `.timeout(publishTimeout)` (+ `TimeoutException` handling) as the initial publish, so a connected-but-silent relay can't wedge the sequential sweep and starve the relays after it. It also no longer skips on the backend's `isAwaitingOk`: that entry lingers after *our* `publishTimeout` fires (the underlying send is abandoned but the backend still counts it awaiting), which would suppress every future resend to that relay. Resends now consult a service-owned in-flight `(relay, event)` set that clears the moment our timeout fires, so a relay that fell silent is eligible again on the next sweep.

Added the exact regression test you described — relay **A** acks immediately, **B** stays silent forever, **C** rejects its first frame then accepts — asserting **C eventually receives the latest event** (and that B doesn't wedge the sweep). Passes.

**@coderabbitai — reconnect fire-and-forget.** `unawaited(_nostrService.reconnectAll())` in `match_control_provider.dart` now has a `.catchError((e) => debugPrint(...))`, so a failed reconnect can't surface as an unhandled async error; the retry/backoff flow is untouched.

**@coderabbitai — scheduling work after disposal.** Added a `_disposed` flag set first in `dispose()`. The unawaited publish attempts no longer schedule a resend timer or reach into the backend after teardown: `_scheduleResendSweep`, the sweep timer body, and `_resendPendingTo` all bail when `_disposed`.

Validation: `flutter analyze` clean on the touched files; full `test/services/nostr` + `test/features/match` suites green (216 tests), including the rust-backed convergence drills.

PR #121, opened 2026-07-19
test: raise hand-written-code coverage to 98.6% + deep review findings

grunch:
## Summary

Deep review of the codebase + a coverage push: **250 new tests** (240 → **490, all green**), taking line coverage of hand-written code from **59.4% → 98.6%** (raw, including generated code: 52.3% → 81.8%). **21 of 27 hand-written files are now at 100%.**

## Coverage policy & tooling

`tool/coverage.sh` runs the suite with coverage and reports through `tool/filter_lcov.dart`, which excludes machine-generated sources (`lib/src/rust/**` flutter_rust_bridge bindings, `lib/l10n/generated/**` gen-l10n output) — covering generated getters proves nothing. Both numbers (filtered and raw) are printed; the policy is documented in the script.

## What got covered

| Area | Before → After |
|---|---|
| create_match_screen | 0.5% → 100% |
| relay_management_screen | 0.5% → 100% |
| relay_config_provider | 1.5% → 100% (connectivity via loopback-only sockets — nothing leaves the machine) |
| submissions_screen | 1.5% → 100% |
| match_providers | 0% → 100% |
| nostr_service | 52% → 100% (NIP-40 expiry, addressable replacement, cache eviction, publish/converge error paths) |
| account_screen | 62% → 99.7% (clipboard, QR, share + failure, import/generate dialogs) |
| home_screen | 42% → 97.6% |
| settings_screen | 63% → 99.6% |
| app_theme / theme & duration providers / key_manager error paths / Filter / models | → 100% |

New support fakes: `RecordingRelayBackend` (nostr_fakes), `relay_fakes` (InMemorySecureStorage, TestableRelayConfigNotifier).

## The 46 lines that remain, and why

- `lib/main.dart` (34): the rust-init bootstrap + `main()` body — `RustLib.init()` can't run under `flutter test` (and double-inits when the rust-tagged suites loaded the .so); constructs real plugins + `runApp`. The testable part (lifecycle observer) is covered.
- `lib/services/nostr/relay/rust_relay_backend.dart` (2): `onError` handlers of the Rust FFI streams — unreachable without forcing the bridge stream itself to error.
- `lib/features/home/home_screen.dart` (6): includes the `_hexToColor` catch, defensively dead (`Match` validation guarantees `#RRGGBB`).
- `lib/features/home/providers/home_providers.dart` (1): private class `_MatchEntry` is **dead code** (never instantiated) — candidate for deletion in a follow-up.
- One-line defensive branches in `match_outcome.dart`, `settings_screen.dart`, `account_screen.dart`.

## Review fix included

**HIGH — `lib/main.dart`**: the bootstrap catch logged the raw `KeyManager.initialize()` exception (`debugPrint('$e')`). That exception derives from stored key material, so its message can carry the key into the device log — and `debugPrint` is not stripped from release builds. Every other path in `key_manager.dart` deliberately avoids this; the catch now logs a generic line only.

## Review findings for follow-up (not fixed here, so this PR stays test-focused)

- **HIGH (bug)**: `testRelayConnectivity` hangs forever for unreachable relays — `await channel?.sink.close()` never completes when the socket never existed, so `addRelay`'s "Adding…" spinner spins indefinitely. Pinned by tests with a documenting timeout wrapper.
- **MEDIUM**: relay URL validation drift — the form accepts `ws://`+`wss://`, the notifier only `wss://` (confusing snackbar); scheme check is case-sensitive while duplicate detection isn't.
- **MEDIUM**: `Filter.search` is silently dropped by `RustRelayBackend.subscribe` (no `search` field in `FilterData`).
- **LOW**: create-match retry re-runs `Match.create` (new ID on retry after partial success); `MatchControlState` lacks `==` (rebuild every timer tick); duplicated `_hexToColor`/`_formatTime` helpers; deprecated `withOpacity` / `Color.red/green/blue` / `Share.share` usages; import-dialog controller retains a typed nsec on barrier dismiss; `match_control_screen.dart` (831) and `match.dart` (806) slightly over the 800-line guideline.

## Test plan
- [x] `flutter test` — 490/490 green (includes the rust-tagged suites against the real crate)
- [x] `flutter analyze` — 0 errors, 0 new warnings from this change
- [x] `./tool/coverage.sh` reproduces the numbers above

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup failure handling by showing a concise, user-friendly initialization message without exposing technical error details.

* **Tests**
  * Expanded coverage for account and key management, match creation and control, scoring and outcomes, home feed behavior, relay settings, submissions, themes, lifecycle handling, and network connectivity.
  * Added coverage tooling to generate filtered reports that focus on application code.

### DIFF
--- a/lib/features/settings/screens/relay_management_screen.dart
+++ b/lib/features/settings/screens/relay_management_screen.dart
@@ -101,8 +101,10 @@ class _RelayManagementScreenState extends ConsumerState<RelayManagementScreen> {
                   if (value == null || value.isEmpty) {
                     return l10n.pleaseEnterRelayUrl;
                   }
-                  if (!value.trim().startsWith('wss://') &&
-                      !value.trim().startsWith('ws://')) {
+                  // Same rule the notifier enforces: secure websockets only.
+                  // Accepting ws:// here would pass the field and then bounce
+                  // off addRelay with a mismatched snackbar.
+                  if (!value.trim().startsWith('wss://')) {
                     return l10n.relayUrlMustStartWithWss;
                   }
                   return null;
--- a/lib/features/settings/settings_screen.dart
+++ b/lib/features/settings/settings_screen.dart
@@ -15,8 +15,11 @@ final packageInfoProvider = FutureProvider<PackageInfo>((ref) {
   return PackageInfo.fromPlatform();
 });
 
-/// Map of supported locales to their display names
-const _localeNames = {
+/// Map of supported locales to their display names.
+///
+/// Public so tests assert against the same source the UI renders from,
+/// instead of duplicating the localized labels.
+const localeDisplayNames = {
   'en': 'English',
   'es': 'Español',
   'pt': 'Português (Brasil)',
@@ -64,7 +67,7 @@ class SettingsScreen extends ConsumerWidget {
               icon: Icons.language,
               title: l10n.language,
               subtitle: currentLocale != null
-                  ? _localeNames[currentLocale.languageCode] ??
+                  ? localeDisplayNames[currentLocale.languageCode] ??
                       currentLocale.languageCode
                   : l10n.systemDefault,
               onTap: () => _showLanguagePicker(context, ref),
@@ -502,7 +505,7 @@ class SettingsScreen extends ConsumerWidget {
             ),
             const Divider(),
             // Language options
-            ..._localeNames.entries.map((entry) {
+            ...localeDisplayNames.entries.map((entry) {
               final isSelected =
                   currentLocale != null && entry.key == currentCode;
               return ListTile(
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -44,8 +44,12 @@ void main() async {
   final keyManager = KeyManager(crypto: crypto);
   try {
     await keyManager.initialize();
-  } catch (e, st) {
-    debugPrint('KeyManager initialization failed: $e\n$st');
+  } catch (_) {
+    // Not logging the exception object: initialize() derives from stored key
+    // material, so its message can carry that material into the device log —
+    // and debugPrint is not stripped from release builds. KeyManager keeps
+    // this invariant everywhere else; this catch must not be the one leak.
+    debugPrint('KeyManager initialization failed');
   }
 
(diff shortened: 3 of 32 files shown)


grunch:
Addressed in 7297dcb.

**@ermeme blockers:**
1. **`tool/coverage.sh` now builds the native crate first** (`cargo build --manifest-path rust/Cargo.toml`, same guard as CI) so the rust-tagged suites can never silently register as skipped — the script now reproduces the advertised number from a clean checkout.
2. **The relay-connectivity hang**: fixed at the root in the stacked **#122** — the failure paths no longer await a `sink.close()` that can never complete, and those two tests drop the `onTimeout` crutch and assert the method returns `false` on its own (test-level timeout fails them if the hang returns). Kept there rather than duplicated here since it's a production behavior change.
3. **`tool/filter_lcov.dart` output path**: always a sibling `lcov.filtered.info` derived from the input's directory — a custom input like `coverage/custom.info` is never overwritten.
4. *(inline)* **Fail-closed**: an empty report (`0/0`) or one missing any hand-written `lib/` source (truncated run, or a file no test imports) now exits non-zero with the missing list, instead of printing an inflated percentage.
5. *(non-blocking)* The tool now prints **both** totals (filtered and raw), matching the PR description.

**@coderabbitai:**
- **Relay form validator** now enforces `wss://` exactly like the notifier — no more field-passes-then-snackbar mismatch; the `ws://` widget test asserts a field error with **no connectivity probe**.
- **Language-picker tests** read labels from the UI's own `localeDisplayNames` map (made public for exactly that) instead of duplicating non-English literals.
- **Submissions failed-save test** restores `SharedPreferencesStorePlatform` via a teardown registered *before* the swap.
- **Resend-failure test** keeps the fake's `connectedRelays` consistent with the reconnect event it emits.
- **Mock events** in the home-providers tests get unique ids (`e-<dTag>-<createdAt>`).

Validation: affected suites green, full suite **490/490**, `flutter analyze` no new issues, and `dart run tool/filter_lcov.dart` reproduces 98.6% / 81.8% with the new fail-closed checks passing.

PR #122, opened 2026-07-19
fix(relays): unreachable relay no longer hangs addRelay forever

grunch:
## Summary

Follow-up to #121, fixing the HIGH bug its review surfaced, plus the dead-code deletion it identified.

**The bug.** `testRelayConnectivity`'s failure paths did `await channel?.sink.close()`. When the handshake never completed (connection refused, silent host), that close future **never resolves** — so the method never returned, `addRelay` never finished, and the "Adding…" spinner spun forever for any unreachable relay.

**The fix.** On the failure paths, close best-effort **without awaiting** (`unawaited(... .catchError(...))`) — there is nothing to wait for on a socket that never existed. The success path (handshake completed) still awaits its close.

**Tests.** The two suites that documented the hang via an `onTimeout` crutch now assert the method returns `false` **on its own** — real regression guards: if the hang comes back, the test-level timeout fails them. (Refused connection returns promptly; the stalled-handshake case returns right after the intended 5s ready-timeout.)

**Dead code.** Deleted the never-instantiated `_MatchEntry` class in `home_providers.dart` (+ its now-unused import) — flagged by the coverage pass and the analyzer; also clears 2 pre-existing analyzer warnings.

## Test plan
- [x] Affected suites green (52 tests), including the two rewritten regression guards
- [x] Full suite: 490/490 green
- [x] `flutter analyze` on changed files: no issues

_Stacked on `test/coverage-100` (#121); GitHub will retarget to `main` when it merges._

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved relay connectivity checks so failed or timed-out connections return promptly instead of hanging while closing.
  - Preserved existing error handling and failure reporting for unavailable or unresponsive relay connections.

- **Tests**
  - Updated connectivity checks to verify reliable failure responses without depending on external timeout workarounds.

### DIFF
--- a/lib/features/home/providers/home_providers.dart
+++ b/lib/features/home/providers/home_providers.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import '../../match/models/match.dart';
-import '../../match/providers/match_providers.dart';
 import '../../../services/nostr/nostr_service.dart';
 
 /// Status filter: which statuses to show on the home screen.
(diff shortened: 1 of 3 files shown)


grunch:
@coderabbitai review

PR #130, opened 2026-07-22
fix: stop shipping a rate-limited relay, and back off when refused

grunch:
Follow-up to #129. That PR fixed a *silent* relay blocking the outbox; this one fixes what the relay probing actually found.

## Measured, not guessed

Bursts of kind 31415 addressable events over a single persistent connection (throwaway key, `expiration` at 1h), the same shape the app publishes:

| Relay | Accepted | Relay message |
|---|---|---|
| `wss://relay.mostro.network` | **5 / 12** | `rate-limited: allowed kinds: max 5 events per minute per IP` |
| `wss://nos.lol` | 25 / 25 | — (~340 ms/event) |
| `wss://relay.primal.net` | 12 / 12 | — |
| `wss://relay.damus.io` | 0 | `status 503`, would not connect |

A match publishes one addressable event per tap. Five is spent in seconds, after which mostro refuses everything and the remote scoreboard stops moving — which is exactly the "events are not real time" symptom.

## Changes

1. **`relay.mostro.network` out of both default lists**, `relay.primal.net` in (`relay_config_provider.dart`, `nostr_service.dart`). It is a Mostro P2P relay; a live scoreboard is not what it is for.
2. **Backoff in the resend sweep.** It retried on a flat 5s heartbeat forever, feeding a rate-limited relay the very requests that keep its limit spent. The interval now doubles per fully-unaccepted sweep, capped at the new `maxResendInterval` (1 min). Any acceptance, or a reconnect, resets it.
3. **Concurrent sweep.** `_resendPendingTo` was awaited per relay in a loop, so a relay that never answers held up every relay behind it for as long as the transport waits for an OK (10s). Relays are swept together now.

## Not changed: `nextCreatedAt` drift

Verified, and I am deliberately leaving it alone — the available fixes are worse than the issue.

`nextCreatedAt` returns `last + 1` on a same-second collision, so a burst can stamp events ahead of the wall clock. Capping the drift would produce *ties*, and NIP-01 breaks a replaceable-event tie by lowest id — so the relay could keep the **older** event, and resending the same event would never win. That is the stale-scoreboard bug the monotonic counter was introduced to prevent.

The exposure is small: publishes are serialised by the outbox and gated on a relay round trip (~340 ms measured), so drift grows ~2 s per second of *sustained* tapping and self-heals the moment the wall clock catches up. Common relay limits are far away (strfry's default `maxTimestampDelta` is 900 s).

## Caveat worth knowing

Defaults only apply to **fresh installs**. Existing installs keep the relay list already in secure storage, so anyone with mostro configured must remove it in Relay Management (or hit reset-to-defaults). A migration that strips it automatically is possible — say the word and I will add it.

## Test plan

- [x] Two new cases in `nostr_service_behavior_test.dart`. Pre-fix they failed with `sweep did not back off: 7 resends in 400ms` and `Actual: ['wss://silent']` (the concurrent sweep never reached the relay behind the silent one).
- [x] Default-relay assertions updated in `nostr_service_behavior_test.dart`, `relay_management_screen_test.dart`, `relay_config_provider_test.dart`.
- [x] `flutter test` — 506 passed.
- [x] `flutter analyze` — 61 issues, unchanged from main; none new.
- [ ] Manual: run a match, watch kind 31415 with `nostreq`, confirm scores land on each press.

## Summary by CodeRabbit

* **Improvements**
  * Updated the default relay selection for new installations to include Primal and Nos.lol.
  * Improved event delivery reliability by retrying missed updates across connected relays.
  * Added adaptive retry timing to reduce repeated attempts when relays are unavailable.
  * Relay reconnections now trigger prompt delivery of pending events.

* **Bug Fixes**
  * Improved handling of relays that reject or fail to receive event updates.
  * Resend attempts now continue independently across relays, preventing one unresponsive relay from blocking others.

### DIFF
--- a/lib/features/settings/providers/relay_config_provider.dart
+++ b/lib/features/settings/providers/relay_config_provider.dart
@@ -62,9 +62,22 @@ class RelayConfig {
 /// Service for managing relay configuration persistence
 class RelayConfigService {
   static const String _relaysKey = 'nostr_relays';
+
+  /// Relays a fresh install starts with.
+  ///
+  /// `relay.mostro.network` used to be here and was removed: it answers
+  /// `rate-limited: allowed kinds: max 5 events per minute per IP`, and a match
+  /// publishes one addressable event per tap — a scoring burst blows through
+  /// five in seconds, after which the relay refuses everything and the remote
+  /// scoreboard stops moving. Measured, not guessed: 5 of 12 events accepted in
+  /// one burst. It is a Mostro P2P relay; a live scoreboard is not what it is
+  /// for.
+  ///
+  /// Existing installs keep whatever is in secure storage, so anyone who
+  /// already has mostro configured has to remove it in Relay Management.
   static const List<String> defaultRelays = [
-    'wss://relay.mostro.network',
     'wss://nos.lol',
+    'wss://relay.primal.net',
   ];
 
   final FlutterSecureStorage _secureStorage;
--- a/lib/features/settings/providers/relay_config_provider.dart
+++ b/lib/features/settings/providers/relay_config_provider.dart
@@ -5,6 +5,8 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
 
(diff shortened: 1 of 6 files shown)


grunch:
@coderabbitai review

PR #131, opened 2026-07-26
feat: sound the match clock — a bell on start, a horn at zero

grunch:
The clock was silent, so a referee had to watch the phone to know when
regulation time ran out. It now says so out loud, twice.

The two cues are deliberate opposites: a short, bright bell strike (G5)
when the fight starts, and a long, low twin air-horn blast (G3) when the
clock reaches zero. Short against sustained, two octaves apart — nobody
has to look at the screen to know which one just fired.

Both are synthesized from scratch by tool/generate_match_sounds.py using
only the Python standard library, so the audio is an original work under
this project's licence: no samples, no third-party rights, no attribution
owed.

Two cases stay silent on purpose. Resuming a paused clock is the same
fight carrying on, and a second bell would read as a second match. And a
match whose time expired while the app was closed is settled without a
horn — it ended long before the notifier existed, and announcing it would
be announcing history.

Sound is a courtesy, never a dependency: the clock and the scoreboard
remain the truth. Missing audio hardware, a denied audio focus, an
unregistered plugin or an uninitialized binding are all logged and
dropped, and MatchControlNotifier defaults to a silent implementation so
nothing that merely constructs one reaches a platform channel.

Linux desktop builds now need GStreamer, which audioplayers links
against — added to the release workflow and the README.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013RAbAVqkFm6nKib182o1bB

## Summary by CodeRabbit

* **New Features**
  * Added an audible bell when a match starts and a horn when regulation ends.
  * Added bundled match sounds with support for Linux and Windows desktop builds.
  * Match scoring remains functional when audio playback is unavailable.

* **Documentation**
  * Documented match sound behavior and updated Linux audio setup instructions.

* **Tests**
  * Added coverage for sound playback timing, silence scenarios, and audio asset validity.


grunch:
@codex review

PR #133, opened 2026-07-26
feat: keep the screen awake while a match is being scored

grunch:
## Problem

A referee's phone locks itself while a fight is on. A minute can pass with nothing to award — a stalled guard, a grip battle — and a phone set to lock after 30 seconds will be asleep by the time the takedown comes. Unlocking a phone to score a point is not a thing that should ever have to happen on the mat.

Video players hold the screen on while they play. A live scoreboard has better reason to.

## What this does

The match control screen holds the screen on for as long as the match could still be scored, and releases it when it could not.

| Match state | Screen held on |
|---|---|
| Waiting to start | yes |
| Running | yes |
| Paused (fighters off the mat, referee conferring) | yes |
| Finished / canceled | no |
| Screen left (back out, navigate away) | no |

Paused counts, deliberately: a stopped clock with the referee conferring is exactly when nothing is being tapped. A decided match releases the hold so a result can sit on the scorer's table without pinning the display on.

No setting, no UI. Nothing to discover and nothing to get wrong — same as a video player.

## How

New `lib/services/wakelock/screen_wakelock.dart`, following the shape `MatchSounds` already established: an abstraction, a no-op for tests, a platform implementation behind a provider.

Two decisions worth calling out:

- **Requests are serialised** through a future chain, not fired as they arrive. A screen opened and closed inside one frame would otherwise apply its two requests in whichever order the platform answered — and losing that race leaves the screen held on for the rest of the session, which nobody notices until the battery is flat.
- **A refused request is rolled back**, not remembered as done. The platform flag needs a foreground activity and there is no guarantee one is there; if we recorded a failed hold as successful, the next release would be skipped as redundant. Either way the failure is logged and dropped — a device that cannot hold its screen on must still referee a match normally, the same rule the match cues follow.

## Platform notes

- **Android** — `FLAG_KEEP_SCREEN_ON` on the activity window. No permission involved; `wakelock_plus` ships an empty manifest, so **nothing new appears on the Play Store listing**.
- **iOS** — the idle timer.
- Neither can hold a phone awake from the background, so a hold left behind cannot flatten a pocketed phone. It is still released explicitly.

## Dependency

`wakelock_plus: ^1.3.3`, held below 1.4.0 on purpose and commented as such in `pubspec.yaml`: from 1.4.0 it wants `win32 ^6`, which `share_plus 10` pins to `^5`, so 1.4.0+ cannot resolve until `share_plus` is upgraded.

`pubspec.lock` adds only the three entries that resolution produces (`wakelock_plus`, `wakelock_plus_platform_interface`, `dbus`) — no existing pin moved. They were inserted by hand rather than by a full re-resolve because the toolchain here (Flutter 3.41.9 / Dart 3.11.5) is a release behind the one that generated the committed lock, and a full `pub get` here rolls 10 unrelated packages back, `audioplayers 6.8.1 → 6.7.1` among them. Worth a second pair of eyes on that file.

## Test plan

- [x] `flutter test` — 547 passing, run three times to rule out flakes
- [x] New coverage: 14 tests across `test/services/wakelock/screen_wakelock_test.dart` and `test/features/match/match_control_wakelock_test.dart`
- [x] Line coverage 100% on `lib/services/wakelock/screen_wakelock.dart` and `lib/features/match/match_control_screen.dart`
- [x] `flutter analyze` — no new issues (61 before, 61 after; all pre-existing `withOpacity` deprecations)
- [x] Verified the Android plugin implementation uses the window flag and declares no permission
- [ ] **Not verified on a physical device** — that a real phone stops locking mid-match is the one claim here that only hardware can confirm

The widget tests cover the case that motivated this: a running match pumped through 60 seconds of no scoring at all, asserting the hold is still in place and was never asked for twice.

## Summary by CodeRabbit

* **New Features**
  * Keeps the device screen awake while a match is in progress, including paused matches.
  * Automatically releases the screen lock when a match finishes or the match screen is closed.
  * Avoids unnecessary screen-lock requests and handles platform failures gracefully.

* **Tests**
  * Added coverage for match lifecycle and screen-lock behavior.

### DIFF
--- a/lib/features/match/match_control_screen.dart
+++ b/lib/features/match/match_control_screen.dart
@@ -1,7 +1,10 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:choke/l10n/generated/app_localizations.dart';
+import '../../services/wakelock/screen_wakelock.dart';
 import '../../shared/theme/app_theme.dart';
 import 'models/match.dart';
 import 'providers/match_control_provider.dart';
@@ -42,6 +45,10 @@ class MatchControlScreen extends ConsumerStatefulWidget {
 }
 
 class _MatchControlScreenState extends ConsumerState<MatchControlScreen> {
+  /// Held onto from [initState] because [dispose] runs once `ref` is no longer
+  /// usable, and releasing the screen is the one thing that must still happen.
+  late final ScreenWakelock _wakelock;
+
   @override
   void initState() {
     super.initState();
@@ -49,6 +56,9 @@ class _MatchControlScreenState extends ConsumerState<MatchControlScreen> {
       DeviceOrientation.landscapeLeft,
       DeviceOrientation.landscapeRight,
     ]);
+
+    _wakelock = ref.read(screenWakelockProvider);
+    _syncWakelock(ref.read(matchControlProvider));
 
     // A match whose clock expired while the app was closed arrives here already
     // waiting: the notifier settles that in its constructor, before this widget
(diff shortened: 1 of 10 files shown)


grunch:
@coderabbitai review

PR #138, opened 2026-07-26
fix: say so when a shared link's pubkey cannot be read

grunch:
Closes #135.

## The problem

`pubkeyFromShareLink` returned `null` for two different situations:

- the link names no board at all, and
- the link names one, and its pubkey is broken.

`openShareLink` treated both as "not for us" and changed nothing. For the first that is right. For the second the user tapped a link meant for one particular board and was left looking at **whatever board they were already watching** — full of real matches, under the wrong identity, with nothing suggesting the link had failed. On a cold start the asynchronous restore supplied that board a moment later, which made it look like the link had worked.

## What changed

The reader now says which of three things it found:

| Case | Meaning | What happens |
|---|---|---|
| `NotAShareLink` | another host, or no pubkey parameter | nothing, silently |
| `SharedBoard` | a readable pubkey | watch it, show the scoreboard |
| `BrokenShareLink` | named a board, pubkey unreadable | say so |

`NotAShareLink` has to stay silent because it is also the ordinary launch route — an app that complained about that would complain every time it opened.

## Why a state and not a snackbar

A message that fades on its own, over somebody else's board, is the same silent substitution this is meant to stop. So the scoreboard renders the broken-link state **instead of** the board, with the pubkey field and the filter chips hidden as well — neither should invite action on a board the user never reached — and clears it only when they press *Show my board*.

The previously watched pubkey is **kept** throughout. It is theirs, and discarding it because a stranger sent a bad link would punish the wrong person. It just waits behind the message.

## Judgement calls worth checking

- **Empty parameter is not broken.** A bare `?npub=` is treated as no parameter, matching the web board's `readSharedPubkey`, because it should not accuse anybody of anything.
- **`openShareLink` now returns true for a broken link.** It reports "this was one of ours", which is what the caller needs; it no longer means "a board was opened".
- **Another host is still ignored entirely**, broken pubkey or not. Not ours to judge.

## Test plan

- [x] `flutter test` — **610 passing**
- [x] Three-case parsing: broken vs absent vs empty vs foreign host vs launch route
- [x] Broken link keeps the stored pubkey and does not open a board
- [x] A good link afterwards clears the state
- [x] The screen hides the board, the input and the chips, and restores them on dismiss
- [x] The rejected value still never reaches the log (an nsec pasted into `?npub=`)
- [ ] Not verified on a device

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated broken-link state for shared scoreboards.
  * Displays a localized explanation when a shared scoreboard link cannot be read.
  * Added a dismiss action that returns to the user’s scoreboard.

* **Bug Fixes**
  * Valid share links continue opening the correct scoreboard.
  * Invalid links are now distinguished from unrelated links and handled appropriately.

* **Documentation**
  * Added broken-link messaging in English, Spanish, Japanese, and Portuguese.

### DIFF
--- a/lib/features/scoreboard/scoreboard_screen.dart
+++ b/lib/features/scoreboard/scoreboard_screen.dart
@@ -3,6 +3,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:choke/l10n/generated/app_localizations.dart';
 
+import '../../services/deep_links/share_link.dart';
 import '../../services/nostr/crypto/nostr_crypto.dart';
 import '../../shared/theme/app_theme.dart';
 import '../../shared/widgets/match_card.dart';
@@ -77,6 +78,7 @@ class _ScoreboardScreenState extends ConsumerState<ScoreboardScreen> {
     final l10n = AppLocalizations.of(context);
     final theme = Theme.of(context);
     final tk = ChokeTokens.of(context);
+    final brokenLink = ref.watch(brokenShareLinkProvider);
     final watched = ref.watch(watchedPubkeyProvider);
     // Two lists: everything in scope, which the chips count from, and what
     // survives the filter, which is what the list shows.
@@ -142,11 +144,12 @@ class _ScoreboardScreenState extends ConsumerState<ScoreboardScreen> {
                 ],
               ),
             ),
-            Padding(
-              padding: const EdgeInsets.fromLTRB(20, 12, 20, 8),
-              child: _buildPubkeyField(l10n, tk, watched),
-            ),
-            if (watched != null)
+            if (!brokenLink)
+              Padding(
+                padding: const EdgeInsets.fromLTRB(20, 12, 20, 8),
+                child: _buildPubkeyField(l10n, tk, watched),
+              ),
+            if (watched != null && !brokenLink)
               Padding(
                 padding: const EdgeInsets.fromLTRB(20, 4, 20, 12),
                 child: StatusFilterBar(
@@ -156,9 +159,12 @@ class _ScoreboardScreenState extends ConsumerState<ScoreboardScreen> {
                 ),
               ),
             Expanded(
-              child: switch ((watched, matches.isEmpty)) {
-                (null, _) => _buildWelcome(l10n, tk),
-                (_, true) => _buildEmpty(l10n, tk),
+              child: switch ((brokenLink, watched, matches.isEmpty)) {
+                // A link that named a board and could not be read outranks
+                // everything: whatever is behind this is not what was tapped.
+                (true, _, _) => _buildBrokenLink(l10n, tk),
+                (_, null, _) => _buildWelcome(l10n, tk),
+                (_, _, true) => _buildEmpty(l10n, tk),
                 _ => _buildList(matches),
               },
             ),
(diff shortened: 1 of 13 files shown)


grunch:
## Adversarial review — PR #138

Reviewed my own work by trying to break it rather than confirm it. Three findings, one of which undoes part of the point of the PR.

---

### 🟠 1. The header still names the old organizer while the body says the link is broken

`lib/features/scoreboard/scoreboard_screen.dart:135-137`

```dart
watched == null
    ? l10n.scoreboardWelcomeTitle
    : _shortPubkey(watched),
```

That subtitle is outside every `brokenLink` guard. So the broken-link state renders like this:

```
┌──────────────────────────────────────┐
│  ▣  Scoreboard                       │
│     3bf0c63f…a2e1        ← still there
├──────────────────────────────────────┤
│            🔗                        │
│      That link is broken             │
│  …the pubkey in it could not be read │
└──────────────────────────────────────┘
```

The screen says "the link is broken" and, one line above, "you are watching `3bf0c63f…a2e1`". A user who does not read carefully concludes that hex **is** the link's board. That is a smaller version of the exact substitution this PR exists to prevent, and it is on screen at the moment the user is most likely to be confused.

The PR body claims the field and the chips are hidden "since neither should invite action on a board the user never reached". The same reasoning applies to the identity in the header, and it was missed.

**Fix:** fall back to the welcome subtitle (or something explicitly about the link) while `brokenLink` is set.

---

### 🟠 2. The host check does not run on a relative URI, so a bare `?npub=` fabricates a broken-link state

`lib/services/deep_links/share_link.dart:67-69`

```dart
if (uri.hasAuthority && uri.host.toLowerCase() != kShareLinkHost) {
  return const NotAShareLink();
}
```

Verified against `Uri`:

```
/?npub=nonsense               -> hasAuthority=false  host=""  npub=nonsense
bjjscore.live/?npub=nonsense  -> hasAuthority=false  host=""  npub=nonsense
https://evil.example/?npub=x  -> hasAuthority=true   host="evil.example"
```

Both of the first two skip the host check entirely — `hasAuthority` is false, so the guard short-circuits before comparing anything. A relative route carrying a `npub` parameter is therefore treated as **ours**, and if the value does not parse it now pops the user onto the scoreboard with an accusatory error about a link they never followed.

This was harmless before this PR (the result was `null`, and null meant "do nothing"). Giving the invalid case a loud, visible state is what turns the gap into a user-facing bug. The change made an existing weakness reachable.

**Fix:** require the host when there is a pubkey parameter to act on, or treat an authority-less URI as `NotAShareLink` unless it is the launch route.

---

### 🟡 3. The log says "ignoring" for the one case that is not ignored

`lib/services/deep_links/share_link.dart:107` — `_logIgnored(uri)` is called on the `BrokenShareLink` branch, which then sets a flag, switches tabs and shows a full-screen message. Whoever reads that log while debugging a support report will conclude the link was dropped.

**Fix:** separate messages, or rename to something honest about both callers.

---

## What held up

- **The three-case split is right and the empty-parameter call is right.** `?npub=` producing `NotAShareLink` matches the web board and avoids accusing anyone over a trailing separator.
- **Keeping the stored pubkey** rather than clearing it. I went looking for a case where a stale board leaks through anyway — finding 1 is that case, but the *data* is correctly untouched.
- **The dismiss-not-snackbar decision** survives scrutiny: a fading toast over another organizer's matches is the failure mode, not a milder version of it.
- **The nsec-in-the-log guard still holds** with the new branching; the test pins it.
- **`openShareLink` returning `true` for a broken link** is right — `didPushRouteInformation` is asking "did you handle this route", not "did you open a board".

## Not covered

No test drives the real cold-start ordering — `_handleLaunchLink`'s post-frame callback racing `WatchedPubkeyNotifier._restore()`. The precedence is correct by construction (the switch checks `brokenLink` first), but the interleaving that motivated the issue is asserted nowhere.

PR #139, opened 2026-07-26
fix: resolve equal-timestamp revisions by NIP-01's id tie-break

grunch:
Closes #136.

## The problem

Two revisions of one match can carry the same `created_at`. That is not exotic — scoring twice inside a second is an ordinary thing for a referee to do. NIP-01 settles it deterministically: **the event with the lowest id wins.**

Nothing did that, and the two places that decided disagreed with each other:

| | Rule it applied | Effect on a tie |
|---|---|---|
| `NostrService._handleIncomingEvent` | `existing.createdAt >= event.createdAt` → drop | **first** to arrive wins |
| `ScoreboardFeedNotifier._upsert` | `createdAt < seen` → drop | **last** to arrive wins |

Relays deliver in whatever order they please, so two phones watching the same board could settle on different revisions of the same match and show **different scores**, with nothing to indicate which was right.

## What changed

One function, `addressableSupersedes`, answers "does this replace what we hold" — newer `created_at` wins; on a tie, lowest id wins — and both callers ask it. Two implementations of "which is newer" is how they stopped agreeing in the first place.

The scoreboard feed had to start keeping the event id to ask the question at all; it tracked only `created_at`. `createdAtOf` still returns what it always did, so the 24-hour age filter above it is unchanged.

## What was deliberately left alone

`MatchFeedNotifier` (home). Its `_upsert` takes a synthetic timestamp from `addLocal` with no event behind it, and that local state is *meant* to outrank relay echoes — the operator's own tap is the truth on their own device. Anything reaching it from the relays has already passed through the service, which now forwards only the winner.

## Test plan

- [x] `flutter test` — **612 passing**
- [x] The rule itself: newer wins, older never wins, tie goes to the lowest id, an event does not replace itself
- [x] **Order independence** — the same pair delivered in opposite orders settles on the same revision
- [x] A genuinely newer revision still wins even when it carries the *higher* id (guards against an id-only rule)
- [ ] Not verified against two real devices on one board

### DIFF
--- a/lib/features/scoreboard/providers/scoreboard_providers.dart
+++ b/lib/features/scoreboard/providers/scoreboard_providers.dart
@@ -142,9 +142,13 @@ class ScoreboardFeedNotifier extends StateNotifier<List<Match>> {
     return '$_idPrefix$short';
   }
 
-  /// Newest event seen per match, so a relay replaying an older revision of an
-  /// addressable event cannot undo a newer one.
-  final Map<String, int> _createdAt = {};
+  /// The event held per match: when it was created, and its id.
+  ///
+  /// The id is here for the tie — two revisions of one match can share a second,
+  /// and NIP-01 settles that by id, not by which relay was quicker. Without it
+  /// this feed accepted whichever arrived last while the service kept whichever
+  /// arrived first, so the two could disagree about the same match.
+  final Map<String, ({int createdAt, String id})> _held = {};
 
   void _onEvent(NostrEvent event) {
     if (event.kind != 31415) return;
@@ -156,16 +160,24 @@ class ScoreboardFeedNotifier extends StateNotifier<List<Match>> {
     if (event.pubkey != _pubkey) return;
 
     try {
-      _upsert(Match.fromNostrEvent(event), event.createdAt);
+      _upsert(Match.fromNostrEvent(event), event.createdAt, event.id);
     } catch (e) {
       debugPrint('Scoreboard: failed to parse event: $e');
     }
   }
 
-  void _upsert(Match match, int createdAt) {
-    final seen = _createdAt[match.id];
-    if (seen != null && createdAt < seen) return;
-    _createdAt[match.id] = createdAt;
+  void _upsert(Match match, int createdAt, String eventId) {
+    final held = _held[match.id];
+    if (held != null &&
+        !addressableSupersedes(
+          arrivingCreatedAt: createdAt,
+          arrivingId: eventId,
+          heldCreatedAt: held.createdAt,
+          heldId: held.id,
+        )) {
+      return;
+    }
+    _held[match.id] = (createdAt: createdAt, id: eventId);
 
     final index = state.indexWhere((m) => m.id == match.id);
     if (index >= 0) {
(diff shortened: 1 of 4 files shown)


grunch:
## Adversarial review — PR #139

Went looking for the ways this could still let two devices disagree, which is the only thing the PR is for.

---

### 🟠 1. The service — where half the bug lived — has no test

`grep 'addressableSupersedes|Ignoring superseded' test/` returns only the pure-function group. Nothing drives an equal-timestamp pair through `NostrService._handleIncomingEvent`.

That is the gap that matters most here. The bug was never "the rule is wrong"; it was **two call sites disagreeing about the rule**, and one of them was the service. This PR changes the service's behaviour in a way nothing observes:

- before: an equal-timestamp event was always dropped, so it was never forwarded to `eventStream`
- after: a lower-id one replaces the cached event **and is forwarded**

So downstream consumers can now receive a second event for the same match in the same second, which they previously never could. `MatchFeedNotifier._upsert` accepts it (it only drops strictly older), so home follows the winner — correct, and entirely untested.

The issue asked for exactly this: *"test both delivery orders for the same `(kind, pubkey, d, created_at)` pair"* — in the service and in the feed. Only the feed got it.

**Fix:** a service-level test pushing the pair through the backend in both orders, asserting the cache and what reaches `eventStream`.

---

### 🟡 2. The comparison is case- and length-sensitive, and the fixtures hide it

`lib/services/nostr/nostr_service.dart` — `arrivingId.compareTo(heldId) < 0`

```
'A'.compareTo('a')      = -1   → uppercase sorts FIRST
'ffff'.compareTo('AAAA') = 1
```

NIP-01 ids are lowercase 64-character hex, so on the wire this is fine. But nothing here normalises or asserts that, and lexicographic order only equals numeric order when both operands share a length and a case. Any id that arrived uppercase — a re-serialisation somewhere, a fixture, a future non-relay source — would order the wrong way, and the failure is the same silent divergence the PR fixes.

The test fixtures make this invisible: they use `'aaaa'` and `'ffff'`, four characters, both lowercase. A rule that broke on real 64-character ids, or on mixed case, would pass every test in this PR.

**Fix:** compare normalised (`toLowerCase()`), and use realistic 64-character ids in at least one test.

---

### 🟡 3. `addressableSupersedes` is public API with no stated contract about ids

It is exported from `nostr_service.dart` and takes bare `String` ids. Nothing in the signature or the doc says they must be same-length lowercase hex, which is the precondition the correctness rests on. The next caller has no way to know.

---

## What held up

- **The rule itself is right**, including the case I most expected to get wrong: a genuinely newer revision carrying a *higher* id still wins. There is a test for that specifically, which is the trap an id-only comparison falls into.
- **Order independence is properly tested in the feed** — the same pair delivered both ways lands on the same revision. That is the assertion that actually encodes the bug.
- **Leaving `MatchFeedNotifier` alone is correct.** I checked whether the local-authoritative path could now be overwritten: `addLocal` stamps `existing + 1` on a same-second write, so it stays ahead of any echo. Touching it would have broken the operator's own tap being the truth on their own device.
- **`createdAtOf` still returns what it did**, so the 24-hour age filter above it is unaffected by the record change.

## Merge-order note

This and #140 both touch `openShareLink`'s neighbourhood, and #138 rewrites the same function. Whichever lands first, the other two need a rebase — worth sequencing rather than merging in parallel.

PR #140, opened 2026-07-26
fix: clear the stack when a shared link arrives

grunch:
Closes #137.

## The problem

`openShareLink` set `watchedPubkeyProvider` and `selectedTabProvider`. Both live **under** the root `Navigator`, so anything pushed on top — `ScoreboardMatchScreen`, `MatchControlScreen` — stayed exactly where it was.

With a match detail open, tapping a shared link:

1. switched the organizer being watched,
2. switched the tab behind the route,
3. left the user looking at the same detail screen.

For a link naming a **different** organizer it was worse than a no-op: the match on screen no longer belonged to the watched pubkey, so `ScoreboardMatchScreen` found nothing for its id and turned into the *"match no longer available"* page. A link that promised a board delivered a dead end.

## What changed

The root navigator is reachable through `navigatorKeyProvider`, and a link pops to the first route before selecting the tab.

**Order matters.** Popping *after* the switch lets the detail screen render once against the new organizer and flash exactly the not-found state this is meant to prevent.

## Why a key and not a router

The app navigates with an `IndexedStack` and imperative pushes. Introducing routing to pop one route would put the riskiest change in the least related place, for a bug that a `popUntil` fixes.

## Behaviour preserved

- **Cold start** pops nothing, having pushed nothing.
- **A link for another host does nothing at all** — worth stating, because an unrelated link must not throw somebody out of the match they are refereeing.

## Test plan

- [x] `flutter test` — **608 passing**
- [x] A pushed route is cleared and the board is what ends up visible
- [x] Cold start still works with an empty stack
- [x] A foreign-host link leaves the stack untouched
- [ ] Not verified on a device — the whole point is warm-start behaviour, which is exactly what a widget test simulates rather than proves

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -188,6 +188,7 @@ class _ChokeAppState extends ConsumerState<ChokeApp>
     final themeMode = ref.watch(themeModeProvider);
 
     return MaterialApp(
+      navigatorKey: ref.watch(navigatorKeyProvider),
       title: 'Choke',
       debugShowCheckedModeBanner: false,
       theme: AppTheme.lightTheme,
--- a/lib/services/deep_links/share_link.dart
+++ b/lib/services/deep_links/share_link.dart
@@ -1,4 +1,4 @@
-import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
 import '../../features/scoreboard/providers/scoreboard_providers.dart';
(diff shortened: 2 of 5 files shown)


grunch:
## Adversarial review — PR #140

One finding here is worse than the bug the PR fixes.

---

### 🔴 1. A shared link now drags a referee out of a running match, past the guard that exists to stop exactly that

`lib/features/match/match_control_screen.dart:155-158`

```dart
return PopScope(
  canPop: !state.isRunning,
  onPopInvokedWithResult: (didPop, _) {
    if (!didPop) _confirmLeave(context);
```

That guard is deliberate: while a match is **in progress**, leaving the control screen requires confirmation.

`popUntil` does not honour it. It calls `Navigator.pop()` in a loop, and `pop()` is unconditional — `PopScope.canPop` is consulted by system back and `maybePop`, not by an imperative `pop`. So:

> A referee is scoring a live match. Someone sends them a board link. They tap it. The match control screen is torn off the stack with no confirmation, and they are looking at somebody else's scoreboard mid-fight.

The match itself survives — its state lives in a provider, not the widget — so nothing is lost, and they can navigate back. But the app's most safety-critical screen just became dismissible by an incoming link, and this PR is what made it so. Trading "a stale detail screen stays visible" for "a live match screen disappears without asking" is a bad trade.

**Fix:** do not pop past a route that refuses to be popped. `maybePop` in a loop respects `PopScope`, or check the running match before popping and leave the stack alone if a match is in progress — the link can wait, or ask.

This case is not covered: the tests push a plain `Scaffold`, which has no `PopScope`, so the guard is never exercised.

---

### 🟠 2. A link for the board already being watched still destroys the user's place

`openShareLink` pops unconditionally on `SharedBoard`, before comparing anything to the current state. Tapping a link for the organizer you are *already* watching — a re-share in a group chat, the same message opened twice — closes the match you had open and returns you to the list, for no change at all.

**Fix:** pop only when the link actually changes the watched board, or when something is on top that would hide it.

---

### 🟡 3. Nothing verifies that `MaterialApp` receives the key

Each navigation test builds its own `MaterialApp` and passes the same `GlobalKey` it overrides the provider with:

```dart
overrides: [navigatorKeyProvider.overrideWithValue(key)],
child: MaterialApp(navigatorKey: key, …)
```

That proves `popUntil` works when the key is wired. It cannot fail if someone deletes `navigatorKey: ref.watch(navigatorKeyProvider)` from `main.dart` — the one line the whole PR depends on in production. Every test would stay green while the feature silently stopped working, which is the same failure shape as the original bug.

---

### 🟡 4. A `GlobalKey` inside a `Provider` is a latent duplicate-key hazard

`navigatorKeyProvider` mints the key in its factory. Nothing invalidates it today, so this is theoretical — but if it were ever refreshed while the old `MaterialApp` was still mounted, two `Navigator`s would briefly claim one `GlobalKey`, which Flutter treats as a framework error rather than a warning.

---

## What held up

- **Popping before switching the tab is the right order**, and for the stated reason: the reverse lets the detail screen render once against the new organizer and flash its own not-found state. That sequencing is the subtle part and it is correct.
- **A foreign-host link leaves the stack alone**, and it is tested. Worth keeping, because it is the difference between "a link I sent works" and "any link anyone sends can interrupt a referee".
- **`currentState?.` tolerates an unattached navigator**, so cold start does not throw.
- **A key instead of a router** remains the proportionate choice for an app that navigates with an `IndexedStack` and imperative pushes.

## Merge-order note

#138 rewrites `openShareLink` and this changes it. They will conflict; sequence them rather than merging in parallel.

PR #141, opened 2026-07-26
feat: paint the board in design 3A when the app is in a light theme

grunch:
Implements design **3A — Broadcast Split, tema claro** from the [Fight Scoreboard design project](https://claude.ai/design/p/20463b59-78fe-4a3a-9301-0148a1af63d6?file=Fight+Scoreboard.dc.html), applied to the full-screen board when the app is in a light theme. The dark theme is unchanged.

## Why this is not a recolour

The board's colours were four private constants and about thirty literals in one file, all chosen against black. Every surface, glow and alpha assumed that background — so swapping only the background is the version of this change that ships **white text on near-white with glows that do nothing**.

They now live in `BoardPalette`, which has exactly two instances, and the screen reads whichever one the app's brightness calls for.

## What 3A specifies, and what it took

| | dark (unchanged) | light (3A) |
|---|---|---|
| surface | `#05070E` | `#F4F6FB` |
| text / labels | white / `#5F6D8A` | `#0D1526` / `#68758F` |
| losing half | `rgba(5,7,14,.62)` | `rgba(244,246,251,.72)` |
| edge & score glow | `.6` | `.32` / `.30` |
| clock card | translucent white, no offset | `#fff`, cast shadow `0 18 46` |
| ADV chip | `#F4B400` / `#FFD451` | `#A16207` / `#854D0E` on `#CA8A04` |
| PEN chip | `#F87171` / `#FCA5A5` | `#DC2626` / `#991B1B` |
| live pill | one green | `#0F9D58` word on `#10B981` surface |

Three of those are the ones worth arguing about:

- **The losing half is veiled in the background itself**, not in black. A dark veil on a light board reads as a shadow rather than as a half that has been washed out.
- **The glows are weaker.** A glow is light in the dark and a shadow in daylight; one alpha cannot be both.
- **The live pill uses two different greens.** The word has to carry contrast, the surface has to stay quiet.

## The part that is easy to miss

A fighter's colour has to be **darkened to 72% before it carries words**. 3A's own default gi colour is `#13C88A` — it glows on black and is illegible as a name on white.

The border and the soft fill keep the raw colour, because that is what the room recognises; only the text is darkened. The dark theme returns it untouched, where the colour the fighter picked is the point.

## Judgement calls

3A does not specify the **waiting, paused or canceled** pills. Rather than invent colours I took them from the palette's own vocabulary: the label grey for a match that has not begun, and the gold and red the chips already speak in.

## Test plan

- [x] `flutter test` — **638 passing**
- [x] The palette switch itself, both directions
- [x] `readable()` takes each channel to 72% on light and is identity on dark
- [x] The veil shares the background's hue in both themes
- [x] Light glows weaker than dark
- [x] A light app paints the board on 3A's surface; a dark app keeps the original
- [x] Fighter names resolve to `#0D1526` and explicitly **not** white — the cheap-version failure this guards against
- [ ] **Not verified on a device.** The ratios and tokens are checked; how it actually reads on a phone in daylight is not.

## Summary by CodeRabbit

* **New Features**
  * Added light and dark visual themes for the full-screen scoreboard.
  * Improved scoreboard styling for cards, statuses, score displays, winner banners, glows, and background washes.
  * Enhanced text readability across different fighter colors and themes.

* **Bug Fixes**
  * Improved scoreboard legibility and color consistency in light and dark modes.

* **Tests**
  * Added coverage for theme-specific colors, readability, and scoreboard rendering.

### DIFF
--- a/lib/features/scoreboard/board_palette.dart
+++ b/lib/features/scoreboard/board_palette.dart
@@ -1,0 +1,211 @@
+import 'package:flutter/material.dart';
+
+/// The colours the full-screen board is painted in.
+///
+/// Two of them. The board is not a normal screen that can take the app's theme
+/// and be done with it: it is a wall display read from across a mat, so every
+/// surface, glow and alpha was chosen against one background. Swapping only the
+/// background leaves white text on near-white and glows that do nothing.
+///
+/// [dark] is the original, which follows choke-scoreboard. [light] is design 3A
+/// — the same structure and the same animations on a light surface, with the
+/// athlete's colour carrying the accent instead of glowing in the dark.
+class BoardPalette {
+  const BoardPalette({
+    required this.background,
+    required this.text,
+    required this.label,
+    required this.loserVeil,
+    required this.cardSurface,
+    required this.cardBorder,
+    required this.cardShadow,
+    required this.cardShadowBlur,
+    required this.cardShadowOffset,
+    required this.vs,
+    required this.advLabel,
+    required this.advValue,
+    required this.advSurface,
+    required this.penLabel,
+    required this.penValue,
+    required this.penSurface,
+    required this.liveText,
+    required this.liveSurface,
+    required this.waiting,
+    required this.paused,
+    required this.canceled,
+    required this.neutralWinner,
+    required this.edgeGlowAlpha,
+    required this.scoreGlowAlpha,
+    required this.dotGlowAlpha,
+    required this.halfWashStrong,
+    required this.halfWashFade,
+    required this.halfWashFadeStop,
+    required this.halfWashEndStop,
+    required this.darkenWinner,
+    required this.backLabel,
+  });
+
+  /// Behind everything.
+  final Color background;
+
+  /// Names, scores, the clock — whatever has to be read first.
+  final Color text;
+
+  /// Column headers and explanations.
+  final Color label;
+
+  /// Laid over the losing half. Carries its own alpha, because it is the
+  /// background veiling what is under it: light on light, dark on dark.
+  final Color loserVeil;
+
+  /// The clock card and the winner banner.
+  final Color cardSurface;
+  final Color cardBorder;
+  final Color cardShadow;
+  final double cardShadowBlur;
+  final Offset cardShadowOffset;
+
+  /// The "VS" under the clock, which is meant to be barely there.
+  final Color vs;
+
+  final Color advLabel;
+  final Color advValue;
+
+  /// The colour the advantage chip's background and border are drawn from.
+  final Color advSurface;
+
+  final Color penLabel;
+  final Color penValue;
+  final Color penSurface;
+
+  /// A running match: the pill's word, and the colour its background and border
+  /// are drawn from — not the same green in the light theme.
+  final Color liveText;
+  final Color liveSurface;
+
+  final Color waiting;
+  final Color paused;
+  final Color canceled;
+
+  /// A finished match nobody won.
+  final Color neutralWinner;
+
+  /// How strongly a fighter's colour glows off the edge bar, the score and the
+  /// name dot. A glow reads as light in the dark and as a shadow in daylight, so
+  /// these are not the same numbers in both themes.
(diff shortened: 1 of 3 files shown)


grunch:
## Adversarial review — PR #141

The PR says "the dark theme is unchanged". It is not, and that is the worst finding here.

---

### 🔴 1. The dark theme's winner banner lost its backdrop

`lib/features/scoreboard/scoreboard_match_screen.dart:575`

On `main` the banner had its own surface, opaque enough to sit on top of anything:

```dart
color: const Color(0xB805070E),   // 72% of the board's own near-black
```

It now draws `palette.cardSurface`, and for the dark palette that is:

```dart
cardSurface: Color(0x08FFFFFF),   // white at 3%
```

So in the dark theme the winner banner is **effectively transparent**. The fighter halves and their colour washes show straight through the panel that announces who won — the one element on the board that a room full of people reads at once.

The mistake was collapsing two different surfaces into one token because they looked alike. They never were: the timer card was always a faint translucent panel (`white .03`, which the token matches exactly) and the banner was always a solid backdrop. 3A gives them the same white only in the *light* theme, and I read that as permission to unify them everywhere.

It also means the border and cast shadow I added for 3A are now on the dark banner, which never had either.

**Fix:** a separate banner surface, and the outline only where 3A asks for it.

---

### 🟠 2. `readable()` darkens colours that were already chosen for the theme

`scoreboard_match_screen.dart:456`

```dart
final word = palette.readable(color);
```

`color` there is whatever the status switch produced — and for four of its five branches that is a palette constant, already correct for the light theme. They get darkened a second time:

| pill | 3A / palette says | rendered |
|---|---|---|
| live | `#0F9D58` | `#0B713F` |
| waiting | `#68758F` | `#4A5467` |
| paused | `#A16207` | `#744605` |
| canceled | `#DC2626` | `#9E1B1B` |

3A is explicit that only one case is darkened — `statusTextL: isFinal ? this.darken(winHex) : '#0f9d58'`. Darkening is for a **fighter's** colour, which arrives from the wire and cannot be trusted to be legible. A colour the palette itself chose needs nothing done to it.

Not a crash, and each is still readable — which is exactly why it would have shipped. The board is just quietly darker than the design in every state except the one that was tested.

---

### 🟡 3. The alphas are the dark theme's in both themes

The colours were tokenised; the opacities they are used at were not.

| | mine (both) | 3A light |
|---|---|---|
| status pill fill | `.12` | `.14` |
| status pill border | `.5` | `.45` |
| chip fill | `.14` | `.12` (ADV) / `.10` (PEN) |
| chip border | `.5` | `.42` |
| winner method chip | `.14` / `.5` | `.14` / `.45` |

Small individually. Together they are the difference between "ported the palette" and "implemented the design", which is what the PR claims to have done.

---

### 🟡 4. The status dot still glows at full strength

`_StatusDot` draws `BoxShadow(color: widget.color, blurRadius: 14)` with no alpha, in both themes. 3A's light pill dot has no shadow at all — a hard 12px circle. A full-opacity blur under a saturated green on `#F4F6FB` reads as a smudge, and it is the one thing on the board that blinks, so it draws the eye to the sloppiest pixel.

---

## What held up

- **The two-palette structure is right**, and the reasoning for it survives: I went looking for a way to do this with `ChokeTokens` and there isn't one — washes, veils and glow alphas are not app chrome.
- **`readable()` itself is correct** where it belongs. The 72% channel scale is what 3A does, and the test pins it against `#13C88A`, the exact colour that motivated it.
- **The veil sharing the background's hue** is the detail I most expected to have got backwards, and it is right in both themes.
- **Weaker glows in daylight** — tested as a relation rather than a magic number, which is the assertion that will survive a future tweak.
- **The "not white" assertion on fighter names** does catch the cheap version of this change. It is the one test here I would keep unchanged.

## What is still not covered

Nothing asserts the *dark* theme's surfaces against their previous values. Finding 1 is exactly the kind of regression a golden test or a handful of value assertions would have caught, and the light-theme tests I added cannot see it by construction.

PR #142, opened 2026-07-26
feat: keep the screen awake while watching a board

grunch:
The control screen already holds the screen for the referee (#133). This is the spectator's side: a board opened from the scoreboard holds it too, from the moment it opens until the viewer leaves.

## Why unconditional, where the referee's is not

The two screens wait differently, and the asymmetry is deliberate:

| | referee (`MatchControlScreen`) | spectator (`ScoreboardMatchScreen`) |
|---|---|---|
| waiting | held — about to start | **held** — waiting for the bell |
| running | held | **held** — a stalled minute is when nobody touches the phone |
| finished | **released** — result sits on the table | **held** — the board is left up to be read across a room |

There is no state in which a spectator's board going dark serves anybody, which is what was asked: *"la pantalla no debe bloquearse en ningún momento si está seleccionada una lucha en el scoreboard, aunque no haya iniciado o finalizado."*

## How

- `initState` takes the hold; `dispose` releases it. The service is read once into a field, because `dispose` runs after `ref` is unusable — same pattern as the control screen.
- **Re-asserted on the clock tick the screen already runs.** A request the platform dropped (no foreground activity yet, a transient refusal) would otherwise stay dropped for the whole match. Repeats are free: the service only touches the platform when held ≠ wanted.
- `ScreenWakelock` itself is unchanged. Its one-owner precondition still holds — this screen and the control screen cannot be mounted together.

## Tests

The scoreboard harnesses gain the `NoopScreenWakelock` override the control screen's tests learned to need the hard way: without it, every board test reaches a real method channel nothing answers, and the 5s timeout becomes a pending timer the framework fails on.

Five new tests: held on open, held while **waiting**, held while **finished** (the deliberate difference from the referee's screen), re-asserted across ticks, released on leave. The recording fake keeps every request including repeats — deduplicating in the fake is how a control-screen test once asserted something unobservable.

## Test plan

- [x] `flutter test` — **647 passing**
- [ ] Not verified on a device — that a real phone stops locking while a board is up is the one claim only hardware can confirm

### DIFF
--- a/lib/features/match/match_control_screen.dart
+++ b/lib/features/match/match_control_screen.dart
@@ -47,7 +47,7 @@ class MatchControlScreen extends ConsumerStatefulWidget {
 class _MatchControlScreenState extends ConsumerState<MatchControlScreen> {
   /// Held onto from [initState] because [dispose] runs once `ref` is no longer
   /// usable, and releasing the screen is the one thing that must still happen.
-  late final ScreenWakelock _wakelock;
+  late final ScreenWakelockLease _wakelock;
 
   @override
   void initState() {
@@ -57,7 +57,7 @@ class _MatchControlScreenState extends ConsumerState<MatchControlScreen> {
       DeviceOrientation.landscapeRight,
     ]);
 
-    _wakelock = ref.read(screenWakelockProvider);
+    _wakelock = ref.read(screenWakelockProvider).lease();
     _syncWakelock(ref.read(matchControlProvider));
 
     // A match whose clock expired while the app was closed arrives here already
@@ -75,7 +75,7 @@ class _MatchControlScreenState extends ConsumerState<MatchControlScreen> {
   @override
   void dispose() {
     SystemChrome.setPreferredOrientations(DeviceOrientation.values);
-    unawaited(_wakelock.keepAwake(false));
+    unawaited(_wakelock.release());
     super.dispose();
   }
 
--- a/lib/features/scoreboard/scoreboard_match_screen.dart
+++ b/lib/features/scoreboard/scoreboard_match_screen.dart
@@ -6,6 +6,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:choke/l10n/generated/app_localizations.dart';
 
 import 'board_palette.dart';
+import '../../services/wakelock/screen_wakelock.dart';
 import '../../shared/widgets/match_card.dart';
 import '../match/models/match.dart';
 import '../match/models/submission_catalog.dart';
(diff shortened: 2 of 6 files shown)


grunch:
## Adversarial review — PR #142

Went after the one claim the commit makes flatly: *"the control screen and this one cannot be up together, since navigating from one to the other tears the first down."* It is false, and the way it is false produces the only interesting findings here.

---

### 🟠 1. A popped board's release can land on top of another screen's hold

Flutter does not dispose a popped route at the moment of the pop — it disposes it **after the exit animation** (~300 ms later). Anything pushed during that window has already taken its hold when the dying board's `dispose` fires `keepAwake(false)`.

Two screens can therefore be "up together" in exactly the sense that matters to a shared boolean:

- **Board → board.** Pop board A, tap another match in the list within the animation. B holds, then A's release lands. B's ticker re-asserts within a second — self-healing, harmless in practice.
- **Board → waiting control screen.** Pop the board, switch tabs, open a *waiting* match in the control screen inside the same window. The control screen holds; the board's stale release lands — and here is the asymmetry this PR forgot: **a waiting control screen never re-asserts.** Verified against the code: its re-assert rides `ref.listen(matchControlProvider…)`, and `MatchControlNotifier` starts no timer for a waiting match (`match_control_provider.dart:126`), so nothing changes until the referee acts. The referee stands at the mat with fighters bowing in, screen free to lock — the exact scenario #133 exists to prevent, reintroduced by a spectator screen they closed seconds earlier.

The second path is humanly narrow (tab switch + tap inside ~300 ms) but it is not impossible, and its failure mode is silent and indefinite.

**Fix options, cheapest first:** re-assert once in the control screen's `initState` post-frame *and* on a periodic tick while waiting (it already ticks while running); or reference-count holds in `ScreenWakelock` so a stale release cannot cancel a newer hold — which is the structural fix the provider's own doc comment already prescribes for a second owner. The doc said "whoever adds one needs reference-counted leases here rather than a bool". This PR added one.

---

### 🟡 2. The "match no longer available" page holds the screen forever

When the match ages out of the feed, `build` returns `_buildGone` — but the ticker keeps running and keeps re-asserting the hold. A dead-end error page now pins the display indefinitely. The instruction was to hold while *a match* is selected; a page whose entire content is "this match is gone" is a stretch of that, and it is also the state most likely to be left face-up on a table overnight.

Cheap fix: stop asserting (or release) when `match == null`; the tick can keep running for the day the match reappears.

---

### 🟡 3. No test drives the gone state's wakelock

The five new tests cover open/waiting/finished/re-assert/leave, but nothing pumps a board whose match is absent and asserts what the wakelock does — which is finding 2's blind spot, and would have surfaced it.

---

## What held up

- **Unconditional hold is the right reading of the ask**, and the waiting/finished cases are tested — including finished, the deliberate divergence from the referee's screen.
- **Re-asserting on the existing ticker** is the correct retry vehicle and costs nothing; the dedup lives in the service where it belongs.
- **The recording fake keeps repeats**, dodging the unobservable-assertion trap the control screen's tests fell into once.
- **The harness `NoopScreenWakelock` override** is in both `_wrap` and `pumpThemed`, and the wakelock tests prove a later override of the same provider wins — the recording fake received calls.
- `late final` + read-in-`initState` correctly outlives `ref` for `dispose`.

## Verdict

Finding 1's second path breaks a safety property of a *different* screen, which is the kind of regression that never shows up in this PR's own tests. It deserves a fix before merge — the post-frame + waiting-tick re-assert is a few lines; the lease refactor is the honest one.

PR #143, opened 2026-07-26
feat: build a Windows binary on release

grunch:
Every release now carries `choke-vX.Y.Z-windows-x64.zip` alongside the APK, the AAB and the Linux tarball, and the release notes tell Windows users what to download and run.

## Shape

A new `build-windows` job on `windows-latest` — it has to be its own runner, since `flutter build windows` needs MSVC and cannot be cross-compiled from the Linux job. It builds, wraps the bundle in a versioned folder, zips it, and hands it to the release job as an artifact; `build-and-release` gains a `needs` edge and a `download-artifact` step before publishing.

## Decisions worth reviewing

- **All-or-nothing on purpose — and it costs latency.** The `needs` edge means a broken Windows build fails the whole release instead of quietly shipping one without it. It also serializes the entire Windows build (~15–20 min) in front of the Android/Linux job, the release and the Play upload. The parallel shape — a third `release` job gating on both builds — means moving four artifacts across jobs, so it is deferred until this file is next touched for structure; for a few-times-a-month release, the all-or-nothing property is worth more than the minutes. If you would rather ship without Windows on failure, that is one `continue-on-error` decision to make explicitly.
- **Tag validation is duplicated, not skipped.** The tag flows into file names and shell in the new job too; a Windows runner is no more entitled to trust a ref than a Linux one.
- **Versioned top-level folder inside the zip**, mirroring the Linux tarball: extraction yields one directory, not a spray of DLLs.
- **No toolchain changes.** Cargokit compiles the Nostr crate during the build; the channel comes from `rust-toolchain.toml` and the MSVC host target is implicit on a Windows host. The cargo cache reuses the existing key pattern, which is already segmented by `runner.os`.

## Test plan

- [x] YAML parses; job graph is `build-windows → build-and-release`
- [x] `windows/` runner scaffold exists in the repo (flutter/runner/CMakeLists.txt)
- [x] `audioplayers_windows`, `flutter_secure_storage_windows` already in the lockfile
- [ ] **Not exercised end-to-end** — workflows only run on a `v*` tag push, so the first real verification is the next release. Worth cutting a throwaway prerelease-style tag on a fork if you want proof before v1.6.5.

## Summary by CodeRabbit

* **New Features**
  * Added Windows x64 build support.
  * Windows installation instructions are now included in release notes.
  * Windows packages are attached to releases alongside Android and Linux downloads.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,88 @@ permissions:
   contents: write
 
 jobs:
+  # Windows cannot be cross-compiled from the Linux runner — flutter build
+  # windows needs MSVC — so it builds on its own runner and hands the archive
+  # to the release job as an artifact.
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Same validation as the release job, and not negotiable here either: the
+      # tag flows into file names and shell, and a Windows runner is no more
+      # entitled to trust a ref than a Linux one.
+      - name: Extract version from tag
+        id: version
+        shell: bash
+        run: |
+          FULL_TAG=${GITHUB_REF#refs/tags/}
+          if ! printf '%s' "$FULL_TAG" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Refusing to build: tag '$FULL_TAG' is not a plain vX.Y.Z tag."
+            exit 1
+          fi
+          echo "full_tag=$FULL_TAG" >> $GITHUB_OUTPUT
+
+      # Cargokit compiles the Nostr crate during flutter build windows; the
+      # channel comes from rust-toolchain.toml and the host target
+      # (x86_64-pc-windows-msvc) is implicit. MSVC itself ships on the runner.
+      - name: Setup Rust
+        run: rustup show
+
+      - name: Cache cargo
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            rust/target
+          key: cargo-release-${{ runner.os }}-${{ hashFiles('rust/Cargo.lock') }}
+          restore-keys: cargo-release-${{ runner.os }}-
+
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          channel: stable
+
+      - name: Install dependencies
+        run: flutter pub get
+
+      - name: Generate localizations
+        run: flutter gen-l10n
+
+      - name: Build Windows bundle
+        shell: bash
+        env:
+          FULL_TAG: ${{ steps.version.outputs.full_tag }}
+        run: |
+          flutter config --enable-windows-desktop
+          flutter build windows --release
+
+      # Wrapped in a versioned top-level directory for the same reason the
+      # Linux tarball is: extraction yields one folder, not a spray of DLLs
+      # into whatever directory the user happened to be in.
+      - name: Package Windows bundle
+        shell: pwsh
+        env:
+          FULL_TAG: ${{ steps.version.outputs.full_tag }}
+        run: |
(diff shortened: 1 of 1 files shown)


grunch:
## Adversarial review — PR #143

Hunted for the ways the Windows binary could differ from the release it ships in. Found one that guarantees it does.

---

### 🔴 1. The Windows exe reports the wrong version — every release, deterministically

The main job does something the new job does not: **it rewrites `pubspec.yaml` before building**.

```yaml
- name: Update version in pubspec.yaml     # main job only
  run: |
    BUILD=$(git rev-list --count HEAD)
    sed -i "s/^version: .*/version: ${VERSION}+${BUILD}/" pubspec.yaml
```

The Android and Linux binaries are built *after* that bump. The Windows job checks out the tag and builds immediately, so its `pubspec.yaml` still holds the **previous** release's version — releasing `v1.6.5`, `choke.exe` self-reports `1.6.4+352`. That version is user-visible: the app reads it through `package_info_plus` and shows it in Account. Every Windows build of every release will claim to be the release before it, and nothing in CI can notice.

**Fix, with a trap inside it:** replicating the bump step is not enough on its own. The build number is `git rev-list --count HEAD`, and the new job's checkout is **shallow** — on a shallow clone that command returns `1`, which would "fix" the version to `1.6.5+1` and break the monotonic build number Android relies on conceptually and Windows should match. The bump replication needs `fetch-depth: 0` on the Windows checkout to count the same history.

---

### 🟠 2. The `needs` edge serializes ~15–20 minutes of Windows build in front of everything

`build-windows → build-and-release` means the Android/Linux job — and the release, and the Play upload — wait for MSVC to compile the Rust crate and the Flutter bundle before starting. The two builds share nothing until the final attach step; they could run in parallel with the release gated on both.

The honest structural fix is a third job (`release`) with `needs: [build-windows, build-and-release]` — but the release step today lives inside the main job and shares its workspace (APK, AAB, tarball, RELEASE_BODY.md), so that refactor moves four artifacts across jobs. Given the release is already the slow path and runs a few times a month, the serialization is defensible **if stated as a cost**, which the PR body does not do — it presents `needs` purely as the all-or-nothing guarantee.

Verdict: acceptable, but the PR should own the latency, and the parallel shape is where this file should head next time it is touched.

---

### 🟡 3. `retention-days: 1` turns a stale re-run into a broken release

If the main job fails after Windows succeeded — signing secret expired, Play upload flake — and someone re-runs the failed job **more than a day later**, `download-artifact` finds nothing and the re-run dies. One day is aggressively short for an artifact whose entire purpose is to survive until a possibly-human-paced retry. The default 90 days is overkill; something like 7 covers a weekend incident.

---

### 🟡 4. Two `channel: stable` resolutions can disagree

Both jobs install "whatever stable is now", independently. A Flutter stable release landing between the two setups gives the platforms different framework versions inside one release. Low probability, real consequence, cheap insurance: pin the version once (workflow-level `env` or `flutter-version:`) and have both jobs read it.

---

## What held up

- **Duplicating the tag validation** instead of skipping it on Windows — the injection reasoning transfers, and the step is byte-identical.
- **The versioned folder inside the zip** mirrors the Linux tarball decision and its rationale.
- **`if-no-files-found: error`** — a missing zip fails loudly at upload instead of silently at attach.
- **No toolchain file changes needed** — verified `rust-toolchain.toml` pins only channel + android targets; the MSVC host target is implicit on a Windows host, and the cargo cache key was already segmented by `runner.os`.
- **All-or-nothing via `needs`** is the right default for the stated reason; finding 2 is about cost disclosure, not the choice.

## Not verifiable here

The workflow only fires on a `v*` tag push. Findings 1 and the packaging path (`build/windows/x64/runner/Release`) are verified by reading, not by a run — the first real execution is the next release, which is late feedback by construction.

grunch:
@coderabbitai review

PR #144, opened 2026-07-27
perf: repaint the board the moment its clock actually changes

grunch:
The board could read **up to ~2 seconds behind the referee's screen while both were computing the same number.** Nothing was late on the wire — the paint was.

## Diagnosis first

Verified there is no polling anywhere on the subscriber path: relay → Rust stream → `NostrService` → feed → provider → screen is push end to end, and the addressable dedup already makes delivery first-relay-wins. Scores arrive as fast as the network allows.

The delay lives in the **clock** — and in both screens at once. It is derived from `startAt` (whole unix seconds), so its true value flips exactly on the second boundary; each screen repaints that flip at its next 1 Hz tick, whose phase is arbitrary. Each display therefore lags the derived truth by **its own tick phase (up to 1 s)**, and the two screens disagree by the difference of their phases **plus device clock skew** — which is how ~2 s happens between a referee screen and a board.

(An earlier revision of this PR claimed truncation and phase stack to 2 s on one screen. They do not — the truncation *is* the flip, and phase measures from it. The review caught it.)

## The fix

- **Boundary-aligned, self-scheduling tickers — on both screens.** The spectator board and `MatchControlNotifier`'s own clock now tick just past the wall-clock second through one shared helper. Aligning only the spectator would pin it to truth while leaving the organizer's display lagging by its press-phase; aligned on both ends, the two displays agree to within device clock skew.
- **Original spectator ticker change.** Each tick schedules the next against the *next* wall-clock second (+40 ms guard, because timers promise "not before", not "exactly at" — a marginally-early fire lands in the old second and paints stale for a full extra one). Re-deriving the delay each tick self-corrects after throttling instead of compounding drift.
- **Selected watch for scores.** The screen watched the whole match list, so on a board where an organizer runs several mats, every other mat's event rebuilt it. It now `select`s its own match's revision and repaints for that alone — a score update never queues behind rebuilds it did not need.

The helper is pure and top-level; tests pin the alignment and that the guard can never schedule into the current second.

## What this deliberately does not claim

- **Publish→relay→fan-out latency for scores** is the organizer's device and the relays; nothing on this screen can shrink it.
- **Wall-clock skew between the two devices** shifts the derived clock by exactly the skew, and no client can correct that without trusting somebody's timestamp.

## Test plan

- [x] `flutter test` — **654 passing**
- [x] `flutter analyze` — 53, unchanged
- [x] `untilNextClockFlip`: wakes just past the boundary; never schedules into the current second
- [x] Existing ticker-dependent tests (clock countdown, wakelock re-vote) unchanged and green
- [ ] The ~2s → sub-second claim is arithmetic, not measurement — worth eyeballing both screens side by side on the next real match

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved match and scoreboard countdown timing to stay aligned with real-world clock second boundaries.
  * Reduced timing drift and avoided stale/delayed second updates.
  * Ensured timers stop cleanly when screens are closed or matches end.
* **Performance**
  * Reduced scoreboard re-rendering by updating only the currently selected match.
* **Tests**
  * Added unit coverage for wall-clock-aligned timing and strengthened scoreboard live-revision behavior checks.

### DIFF
--- a/lib/features/match/providers/match_control_provider.dart
+++ b/lib/features/match/providers/match_control_provider.dart
@@ -4,6 +4,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import '../models/match.dart';
 import '../models/match_outcome.dart';
 import '../../../services/audio/match_sounds.dart';
+import '../../../shared/wall_clock.dart';
 import '../../../services/nostr/nostr_service.dart';
 import '../../home/providers/home_providers.dart';
 
@@ -441,17 +442,29 @@ class MatchControlNotifier extends StateNotifier<MatchControlState> {
     _publishState();
   }
 
+  /// Tick on the wall-clock second, not on a phase set by when the referee
+  /// happened to press start. The clock is derived from whole-second startAt,
+  /// so its truth flips on the boundary — and a display that repaints there
+  /// agrees with every other boundary-aligned display of the same match (the
+  /// spectator board), instead of each lagging truth by its own arbitrary
+  /// phase. Self-scheduling also self-corrects after a throttled tick.
   void _startTimer() {
     _timer?.cancel();
-    _timer = Timer.periodic(const Duration(seconds: 1), (_) {
-      final remaining = _calculateRemaining(state.match);
-      state = state.copyWith(remainingSeconds: remaining);
+    _timer = Timer(
+      untilNextClockFlip(DateTime.now().millisecondsSinceEpoch),
+      () {
+        if (!mounted) return;
+        final remaining = _calculateRemaining(state.match);
+        state = state.copyWith(remainingSeconds: remaining);
 
-      // Regulation time is over.
-      if (remaining <= 0) {
-        _onTimeUp();
-      }
-    });
+        // Regulation time is over.
+        if (remaining <= 0) {
+          _onTimeUp();
+          return;
+        }
+        _startTimer();
+      },
+    );
   }
 
   /// Queue the current match state for publishing to Nostr.
--- a/lib/features/scoreboard/scoreboard_match_screen.dart
+++ b/lib/features/scoreboard/scoreboard_match_screen.dart
@@ -128,13 +128,15 @@ class _ScoreboardMatchScreenState extends ConsumerState<ScoreboardMatchScreen> {
   @override
   Widget build(BuildContext context) {
     final l10n = AppLocalizations.of(context);
-    // Selected, not watched whole: a busy organizer runs several mats, and
-    // every other mat's event rebuilds whoever watches the full list. This
-    // screen repaints for its own match's revisions only, so a score update
-    // never queues behind rebuilds it did not need.
-    final match = ref.watch(scoreboardMatchesProvider.select(
-      (matches) => matches.where((m) => m.id == widget.matchId).firstOrNull,
-    ));
+    // Watched whole, deliberately not select()ed. A select here compares the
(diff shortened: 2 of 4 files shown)


grunch:
## Adversarial review — PR #144

Went after the numbers, since the whole PR is a latency claim. The mechanism survives; the arithmetic does not.

---

### 🟠 1. The "~2s = truncation + phase" accounting double-counts, and the PR ships it in three places

The commit, the PR table and the code comment all stack "up to 1s lost to truncating now" **on top of** "up to 1s of ticker phase". That is not how the two compose. Truncated `now` changes exactly at the wall-second boundary; a screen repaints that change at its next tick, so its display lags **truth** by its tick phase alone — the truncation *is* the flip, and the phase already measures from it. Spectator-side recoverable lag: **≤1 s worst, ~0.5 s expected.** Not two.

Where the user's ~2 s plausibly comes from is the *pair* of screens: the organizer's own display has the same defect this PR fixes — `MatchControlNotifier._startTimer` is `Timer.periodic(1s)` phased to the moment the referee pressed start (verified, `match_control_provider.dart:453`) — plus wall-clock skew between two devices that are not NTP-agreeing. Display-to-display mismatch = |orgPhase − spPhase + skew|, and ~1 s of skew plus adverse phases reads as ~2 s.

Which exposes the uncomfortable corollary: **pinning the spectator to the boundary optimizes distance-to-truth, not distance-to-the-organizer's-display.** Against the organizer's screen, expected mismatch moves from E|p₁−p₂| ≈ 0.33 s to E|orgPhase − 0.04| ≈ 0.46 s — *slightly worse on average* — while the worst case improves (~1 s → ~0.96 s… i.e., barely) and the spectator becomes exactly right in absolute terms. The defensible position is "paint the true clock, and fix the organizer's phase the same way in a follow-up, at which point both screens converge to within milliseconds plus skew". The PR does not say that; it implies this change alone closes the gap.

**Fix:** correct the arithmetic in the comment and the PR body, and name the follow-up (aligning `_startTimer` the same way) instead of overclaiming this half.

---

### 🟡 2. A throttled tick now skips seconds silently

Self-scheduling from reality means a tick delivered late paints the *current* truncated second — correct, but the display can jump 05:00 → 04:58 with no intermediate paint. `Timer.periodic` had the same failure with different symptoms (burst of stale ticks). Correct trade, worth one sentence in the comment so the next reader knows the jump is chosen, not a bug.

### 🟡 3. The guard-bound test hardcodes the default

`lessThanOrEqualTo(1040)` re-states `guardMs: 40` numerically; changing the default breaks the test without breaking behavior. Derive the bound from the parameter or pass it explicitly.

---

## What held up

- **The diagnosis discipline**: polling was hunted for and ruled out before anything was "optimized" — the score path is genuinely push end to end, and saying "nothing was late on the wire" is correct.
- **The guard's reason is right**: timers promise "not before", and a marginally-early fire would paint stale for a full second. The pure-helper test that no schedule can land inside the current second pins exactly that.
- **`select` on the match instance** is sound: `_upsert` replaces the instance only for the updated match, so identity comparison rebuilds precisely when this match's revision changes. Verified against the provider's sort producing new lists — the selected value survives them.
- **Dispose/leak safety** of the recursive chain rides on the framework's own pending-timer assertion, which every widget test here exercises implicitly.
- **The honest "what this does not claim" section** — skew and fan-out are correctly declared out of reach. The problem is the claimed size of what *is* in reach, not the boundary.

grunch:
@coderabbitai review

PR #153, opened 2026-07-31
refactor: read a shared link's key from npub, and only npub

grunch:
Removes the `pubkey` query-parameter alias. A shared board link is read from `?npub=` and nothing else.

## Why now, and why it is safe

This is a **breaking change to a public URL contract**, and the only reason it can be made is that the contract was never exercised:

- `liveBoardShareUrl` produces `?npub=`.
- choke-scoreboard's `buildShareLink` sets `npub`.
- The web commit that introduced the alias, `1a294dd`, is titled *"auto-load an organizer from a shared link **(?npub=)**"*.

Nothing in either repo has ever generated a `?pubkey=` link, so there are none in the wild to break. The residual risk is a third party who hand-built one against an alias documented nowhere but in a code comment.

The window for this is now. A shared link lives in a chat forever — once real `?pubkey=` links exist, removing support stops being reversible.

## Why remove it at all

Two names for one thing doubles the cases to cover and the sentences needed to explain the URL, and it leaves "which one is canonical?" to be asked every time somebody builds a link. It also came up as the counter-example while specifying match links (protolayer-io/choke#152 §2.3): the reason *not* to add a second name for `match` is this pair, which exists as historical drift rather than because two names were ever wanted.

## Changes

- `kSharePubkeyParams` (list) → `kSharePubkeyParam` (single const). `readShareLink` no longer loops; its three outcomes are unchanged.
- `_log` reports the one parameter name it now knows about.
- Doc comments corrected, including the claim that the app "accepts exactly what the web board accepts".
- `docs/deep-links.md` records the removal and why, so the next reader does not re-add it.

## Test plan

- [x] `flutter test` — 669 passing
- [x] `flutter analyze` — 0 errors
- [x] The test that asserted the alias was **inverted, not deleted**: `?pubkey=…` now asserts `NotAShareLink`. Silently accepting it again is the regression worth catching.
- [x] Confirmed no remaining references — the surviving `'pubkey'` hits are the Nostr event field in `nostr_service.dart`, unrelated

Pairs with protolayer-io/choke-scoreboard#35, which does the same on the web side.

### DIFF
--- a/docs/deep-links.md
+++ b/docs/deep-links.md
@@ -26,10 +26,14 @@ Both need the platform to hand the link to the framework rather than merely
 launching the activity, which is what `flutter_deeplinking_enabled` does in the
 Android manifest.
 
-The accepted parameters are `npub` and `pubkey`, holding either an `npub1…` or
-64-character hex — the same pair choke-scoreboard accepts (`SHARE_PUBKEY_PARAMS`
-in its `share-link.ts`), because the same URL is opened by whichever of the two
-the recipient has.
+The accepted parameter is `npub`, holding either an `npub1…` or 64-character hex
+— the same name choke-scoreboard accepts (`SHARE_PUBKEY_PARAM` in its
+`share-link.ts`), because the same URL is opened by whichever of the two the
+recipient has.
+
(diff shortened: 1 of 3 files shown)


grunch:
@coderabbitai review

PR #154, opened 2026-07-31
feat: read a link that names one match, not just a board

grunch:
Step 1–2 of the build order in `docs/specs/shared-match-links.md` (§9.3): the parser and the link handler. **Nothing shows a match yet, and nothing produces such a link.** That is the point — §9.2 requires readers to land before writers, so the links that eventually go out are readable the day they go out.

## What this adds

- `kShareMatchParam` (`match`) and `matchShareUrl(npub, matchId)`.
- A fourth `ShareLink` case, `SharedMatch(pubkeyHex, matchId)` — both halves always, because an id alone names nothing.
- `requestedMatchProvider`: what a link asked for, until something shows it. `openShareLink` records the request; it does not resolve it. Pending / Resolved / Unresolved are states of a screen, not of a URL, and the screen is the next change.

## The parsing rules, as specified

A match id is four lowercase hex characters — what `Match.create` generates. Validation runs on the **decoded** value, trimmed and lowercased, because both platforms percent-decode before this code sees anything and reading the raw query to get underneath that would be fighting the platform for nothing.

| Link | Result |
|---|---|
| `?npub=X&match=abcd` | `SharedMatch` |
| `?npub=X&match=%61%62%63%64` / `ABCD` / `abcd%20` | `SharedMatch` — same id |
| `?npub=X&match=abc` / `abcde` / `zzzz` / `ab-cd` | **Broken** |
| `?npub=X&match=` | `SharedBoard` — empty is absent |
| `?match=abcd` (no npub) | **Broken** |

Accepting more than we emit is deliberate: chat clients and auto-capitalising keyboards mangle what people paste, and the `npub` reader already trims for the same reason.

Two cases are worth their own sentence. A **malformed id is broken, not absent** — the sender named something, and quietly handing over the board it sits on is exactly the substitution this file exists to refuse. And an **id with no organizer is broken** because it cannot work: ids are unique only inside one author's events, so there is nobody to look it up under.

## Behaviour change, per §6.1

The "only clear the stack when there is something to gain" rule widens. A re-share of the match already on screen still changes nothing — that is the group-chat case and it reads as the app losing the viewer's place. But a link naming a *different* match on the board already being watched now has something to do, where a board link had nothing.

The referee protection of §6.2 is untouched: `_clearStack` still stops at the first route that refuses to go.

## Test plan

- [x] `flutter test` — 679 passing, 11 new cases covering the table above, the stack rule, and that a board or broken link clears any outstanding request
- [x] `flutter analyze` — 0 errors
- [x] Existing deep-link tests extended rather than rewritten around, per §9.3

Pairs with the web reader in choke-scoreboard, which lands alongside this per §9.2.

## Summary by CodeRabbit

- **New Features**
  - Added shareable links that open a specific match directly.
  - Match links now validate match identifiers and handle navigation to the requested match.
  - Link handling distinguishes board, match, broken, and unrelated links.

- **Bug Fixes**
  - Improved broken-link messages across supported languages to describe unreadable link content more accurately.

- **Documentation**
  - Clarified that match links can be read and recorded but are not yet displayed or generated.

### DIFF
--- a/docs/deep-links.md
+++ b/docs/deep-links.md
@@ -1,9 +1,12 @@
 # Shared board links
 
 > **A link to one particular match** — `?npub=…&match=…` — is specified in
-> [`specs/shared-match-links.md`](specs/shared-match-links.md) and is not built
-> yet. Everything below describes the board link, which that one extends rather
-> than replaces.
+> [`specs/shared-match-links.md`](specs/shared-match-links.md). It is **read**
+> but not yet **shown**: `readShareLink` returns `SharedMatch` and
+> `openShareLink` records the request in `requestedMatchProvider`, and the
+> scoreboard does not act on it yet. Nothing produces such a link either.
+> Everything below describes the board link, which that one extends rather than
+> replaces.
 
 Sharing a pubkey from Account produces an ordinary web link:
 
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -882,9 +882,9 @@
   "@scoreboardBrokenLinkTitle": {
     "description": "Title shown when a shared link named a board but its pubkey could not be read"
   },
-  "scoreboardBrokenLinkBody": "It was meant for a specific board, but the pubkey in it could not be read. Ask whoever shared it to send it again.",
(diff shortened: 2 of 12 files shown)


grunch:
@coderabbitai review

